PR #3, opened 2021-09-02
Patch 1

rajeevsharmaCS:


tivv (changes requested):
Please provide description on the purpose of this builds. Note that this repo should not be running tests. Tests should be running in the respective repos of plugins, e.g. see https://github.com/data-integrations/google-cloud/blob/e2e_poc/.github/workflows/e2e.yml.
We still need to run tests for the framework itself (if any) in this repo, but please follow the approach in https://github.com/cdapio/cdap

PR #7, opened 2021-09-10
All the executables related to OS

ksahilCF:
Window, Mac and Linux executables

### DIFF
--- a/src/main/java/utils/SeleniumDriver.java
+++ b/src/main/java/utils/SeleniumDriver.java
@@ -3,6 +3,7 @@ package utils;
 import org.openqa.selenium.By;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.chrome.ChromeDriver;
+import org.openqa.selenium.chrome.ChromeOptions;
 import org.openqa.selenium.remote.HttpCommandExecutor;
 import org.openqa.selenium.remote.SessionId;
 import org.openqa.selenium.support.ui.ExpectedConditions;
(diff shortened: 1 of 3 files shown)


src/main/java/utils/SeleniumDriver.java
```diff
@@ -24,11 +25,37 @@
...
+    public static String chromeWindow=System.getProperty("user.dir")+ ("//executables/window/chromedriver.exe");
+    public static String chromeMac=System.getProperty("user.dir")+ ("//executables/mac/chromedriver");
+    public static String currentOS;
 
       SeleniumDriver() {
-        System.setProperty("webdriver.chrome.driver", "./src/test/resources/webdriverFiles/chromedriver");
-        // driver = new ChromeDriver();
-        chromeDriver=new ChromeDriver();
+          currentOS=System.getProperty("os.name");
+          System.out.println("++++++++"+currentOS);
```
tivv: Please remove or use proper logging

src/main/java/utils/SeleniumDriver.java
```diff
@@ -24,11 +25,37 @@
     public final static int TIMEOUT = 50;
     public final static int PAGE_LOAD_TIMEOUT = 50;
     public static URL url;
+    public static String chromeLinux=System.getProperty("user.dir")+ ("//executables/linux/chromedriver");
```
tivv: Why is it taken from user dir?

src/main/java/utils/SeleniumDriver.java
```diff
@@ -24,11 +25,37 @@
...
+    public static String currentOS;
 
       SeleniumDriver() {
-        System.setProperty("webdriver.chrome.driver", "./src/test/resources/webdriverFiles/chromedriver");
-        // driver = new ChromeDriver();
-        chromeDriver=new ChromeDriver();
+          currentOS=System.getProperty("os.name");
+          System.out.println("++++++++"+currentOS);
+          if(currentOS.toLowerCase().contains("mac")) {
+              System.setProperty("webdriver.chrome.driver", chromeMac);
```
tivv: Please check if property is already set. In this case don't overwrite it. 
BTW: Where are this executables come from? What is their license? Why are we including then as binaries?

PR #9, opened 2021-09-10
Project id is cleared and then overriden

ksahilCF:


tivv (commented):
Why are you dropping chromedriver in this PR?

PR #12, opened 2021-09-14
Project id is cleared before overridden

ksahilCF:
Project Name is cleared

### DIFF
--- a/src/main/java/pages/actions/CdfGcsActions.java
+++ b/src/main/java/pages/actions/CdfGcsActions.java
@@ -21,11 +21,8 @@ public class CdfGcsActions {
         CdfGCSLocators.referenceName.sendKeys(UUID.randomUUID().toString());
     }
 
-    public static void enterProjectId()  {
-        for (int i = 0; i <=100; i++) {
-            CdfGCSLocators.projectID.sendKeys(Keys.BACK_SPACE);
-        }
-        CdfGCSLocators.projectID.sendKeys("cdf-athena");
+    public static void enterProjectId() {
+        SeleniumHelper.replaceElementValue(CdfGCSLocators.projectID,"cdf-athena");
     }
 
     public static void enterGcsBucket(String bucket) throws IOException {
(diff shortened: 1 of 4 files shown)


src/main/java/pages/actions/CdfGcsActions.java
```diff
@@ -20,6 +22,9 @@ public static void enterReferenceName()  {
     }
 
     public static void enterProjectId()  {
+        for (int i = 0; i <=100; i++) {
```
tivv: Why are we doing it?

tivv: I think it's better just to call `clear()`

ksahilCF: I have tried with that but the issue is clear doesn't work on reactive or intelligent fields. This is the workaround via keyboard. It is also logged as a web driver issue. Please follow the link - https://github.com/SeleniumHQ/selenium/issues/6741

ksahilCF: Other work around-WebElement ele = webDriver.findElement(By.xpath(".//input[@name='username']"));
      new Actions(webDriver).click(ele)
        .pause(200).keyDown(Keys.CONTROL).sendKeys("a").keyUp(Keys.CONTROL)
        .pause(200).sendKeys(Keys.BACK_SPACE)
        .pause(200).sendKeys("KOOKOK").perform();

I believe for look if fast to execute that pausing for performing

tivv: Could you please extract it into the utility class and method and add a javadoc with the link above? This would explain it.
I may make sense to pass new value as well, this way its would look similar to `util.ReplaceElementValue(element, 'value')`

ksahilCF: Added method replaceElementvalue and changed the scope of Junit dependency to compiletime. Project value is hardcoded as there is separate PR. Once you will approve this will raise that change in the other PR

ksahilCF: java doc added

tivv (approved):

PR #16, opened 2021-09-22
screen resolution increased failing testcase

ksahilCF:
Test case will still fail as Bigquery client will be integrated in next PR

### DIFF
--- a/src/main/java/utils/SeleniumDriver.java
+++ b/src/main/java/utils/SeleniumDriver.java
@@ -32,6 +32,7 @@ public class SeleniumDriver {
         ChromeOptions chromeOptions = new ChromeOptions();
         chromeOptions.addArguments("--no-sandbox");
         chromeOptions.addArguments("--headless");
+        chromeOptions.addArguments("window-size=1920x1040");
         chromeOptions.addArguments("--disable-gpu");
         chromeDriver= new ChromeDriver(chromeOptions);
         chromeDriver.manage().window().maximize();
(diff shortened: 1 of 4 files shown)


src/main/java/utils/SeleniumDriver.java
```diff
@@ -32,6 +32,7 @@
         ChromeOptions chromeOptions = new ChromeOptions();
         chromeOptions.addArguments("--no-sandbox");
         chromeOptions.addArguments("--headless");
+        chromeOptions.addArguments("window-size=1920x1040");
```
tivv: It should be `--window-size`, also can we put it into config?

ksahilCF: In Java it works that way-https://www.tabnine.com/code/java/methods/org.openqa.selenium.chrome.ChromeOptions/addArguments

No need to put it config as this will be base class and it doesn't matter if we are running in headless or browser this can be constant.

tivv: According to the [javadoc](https://www.selenium.dev/selenium/docs/api/java/org/openqa/selenium/chromium/ChromiumOptions.html#addArguments(java.util.List)), both are working, so it's better to use consistent style. 
I understand that it can be same for headless or not, but we may want to change it at some point going forward, so having it configurable would be a plus

src/test/resources/ConnectionParameters.properties
```diff
@@ -1,9 +1,7 @@
 DownloadPath=/Users/ksahil/Downloads/
 
-#My username and Passwords
-gLoginUsername=ksahil@google.com
-gPassword=*********
-
+Project-ID=cdf-athena
+windowSize=--window-size=1920x1040
```
tivv: Can you just have `1920x1040` here and have `--window-size=`  in the code?

ksahilCF: Done

tivv (approved):

PR #17, opened 2021-09-22
BigQuery Client integration with test case

ksahilCF:
Test Case should pass completely. No of records are fetched from BQ client

### DIFF
--- a/src/main/java/utils/GcpClient.java
+++ b/src/main/java/utils/GcpClient.java
@@ -9,7 +9,6 @@ public class GcpClient {
 
         String projectId = SeleniumHelper.readParameters("Project-ID");
         String datasetName = SeleniumHelper.readParameters("Data-Set");
-        ;
         String tableName = SeleniumHelper.readParameters(table);
         String query = "SELECT count(*) "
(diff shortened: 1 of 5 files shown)


src/main/java/utils/GcpClient.java
```diff
@@ -0,0 +1,55 @@
...
+        ;
+        String tableName = SeleniumHelper.readParameters(table);
+        String query = "DROP TABLE `"
+                + projectId
+                + "."
+                + datasetName
+                + "."
+                + tableName
+                + "`";
+
```
tivv, line 37: Do you want to run the query?

ksahilCF: yes, i want to get the count of records transferred

tivv: So I don't see where you run the query. I see only where you make the query string

ksahilCF: just check the return statement it is calling the query function executing the string

tivv: There is no return statement in this method

PR #22, opened 2021-10-01
Changes to GCP client for specific project

ksahilCF:


ksahilCF:
@rmstar @tivv please review the GCPCLient PR

src/main/java/io/cdap/e2e/utils/GcpClient.java
```diff
@@ -20,7 +20,8 @@ public static int countBqQuery(String table) throws IOException, InterruptedExce
                 + "."
                 + table
                 + "`";
-        BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();
+        BigQuery bigquery =  BigQueryOptions.newBuilder().setProjectId(SeleniumHelper.readParameters("Project-ID"))
```
tivv: Do same for `dropBqQuery` and extract common code in a separate method

ksahilCF:
@rmstar @tivv please review this common method PR

src/main/java/io/cdap/e2e/utils/GcpClient.java
```diff
@@ -6,24 +6,25 @@
 import com.google.cloud.bigquery.TableResult;
 import java.io.IOException;
 
+
 public class GcpClient {
+    public static  String query;
```
tivv: Never use volatile static fields. It's a big anti-pattern. In this case you want `createBigQueryClient()` just to return properly configured `BigQuery` object and `countBqQuery`/`dropBqQuery` can simply use it to run necessary query

src/main/java/io/cdap/e2e/utils/GcpClient.java
```diff
@@ -8,45 +8,47 @@
...
 
-    public static void createBigQueryClient() throws InterruptedException, IOException {
+    public  void createBigQueryClient(String query) throws InterruptedException, IOException {
         BigQuery bigquery =  BigQueryOptions.newBuilder().setProjectId(SeleniumHelper.readParameters("Project-ID"))
                 .build().getService();
         QueryJobConfiguration queryConfig = QueryJobConfiguration.newBuilder(query).build();
        TableResult results = bigquery.query(queryConfig);
-        noOfRecords=Integer.parseInt((String)results.getValues().iterator().next().get(0).getValue());
+       if (results.getTotalRows()>0) {
+           noOfRecords = Integer.parseInt((String) results.getValues().iterator().next().get(0).getValue());
```
tivv: Instead of having it as a variable, just return it. It's always better to have less mutable state

src/main/java/io/cdap/e2e/utils/GcpClient.java
```diff
@@ -8,45 +8,47 @@
...
                 + datasetName
                 + "."
                 + table
                 + "`";
-        createBigQueryClient();
+        createBigQueryClient(dropQuery);
     }
 
-    public static void createBigQueryClient() throws InterruptedException, IOException {
+    public  void createBigQueryClient(String query) throws InterruptedException, IOException {
```
tivv: As soon as you are not only creating a client, but also executing a query, name the method appropriately (e.g. `executeQuery`).

src/main/java/io/cdap/e2e/utils/GcpClient.java
sahilkapoor301992: returned  directly.

src/main/java/io/cdap/e2e/utils/GcpClient.java
sahilkapoor301992: changed method to execute query

src/main/java/io/cdap/e2e/utils/GcpClient.java
```diff
@@ -6,43 +6,47 @@
...
+    public int executeQuery(String query) throws InterruptedException, IOException {
+        BigQuery bigquery =  BigQueryOptions.newBuilder().setProjectId(SeleniumHelper.readParameters("Project-ID"))
+                .build().getService();
         QueryJobConfiguration queryConfig = QueryJobConfiguration.newBuilder(query).build();
-        TableResult results = bigquery.query(queryConfig);
-
-
+       TableResult results = bigquery.query(queryConfig);
+       if (results.getTotalRows()>0) {
+         return Integer.parseInt((String) results.getValues().iterator().next().get(0).getValue());
```
rmstar: what does this method return? Can you add a comment to clarify?

src/main/java/io/cdap/e2e/utils/GcpClient.java
```diff
@@ -6,43 +6,47 @@
...
                 + "."
                 + datasetName
                 + "."
                 + table
                 + "`";
+        executeQuery(dropQuery);
+    }
 
-        BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();
+    public int executeQuery(String query) throws InterruptedException, IOException {
```
rmstar: this can be private

src/main/java/io/cdap/e2e/utils/GcpClient.java
tivv: Also I believe instead of doing `getValue` and parsing you can just call `getLongValue`

src/main/java/io/cdap/e2e/utils/GcpClient.java
sahilkapoor301992: Made private

src/main/java/io/cdap/e2e/utils/GcpClient.java
sahilkapoor301992: Added comment and no of records present should be int so parsed it in integer

tivv (approved):

PR #48, opened 2021-12-06
BigQuery and GCS Plugin related common changes

sawantpritam:
@tivv @rmstar @itsankit-google Please review.

### DIFF
--- a/src/main/java/io/cdap/e2e/pages/actions/CdfStudioActions.java
+++ b/src/main/java/io/cdap/e2e/pages/actions/CdfStudioActions.java
@@ -17,8 +17,10 @@
 package io.cdap.e2e.pages.actions;
 
 import io.cdap.e2e.pages.locators.CdfStudioLocators;
+import io.cdap.e2e.utils.ConstantsUtil;
 import io.cdap.e2e.utils.SeleniumDriver;
 import io.cdap.e2e.utils.SeleniumHelper;
+import org.openqa.selenium.By;
 import org.openqa.selenium.JavascriptExecutor;
 import org.openqa.selenium.WebElement;
 
@@ -80,5 +82,35 @@ public class CdfStudioActions {
 
   public static void clickSink() {
     CdfStudioLocators.sink.click();
+  }
+
+  public static void clickValidateButton() {
+    CdfStudioLocators.validateButton.click();
+  }
+
+  public static void clickProperties(String plugin) {
+    SeleniumDriver.getDriver().findElement(
+      By.xpath("//*[contains(@title,'" + plugin + "')]//following-sibling::div")).click();
+  }
+
+  public static void  connectSourceAndSink(String source, String sink) {
+    SeleniumHelper.waitElementIsVisible(SeleniumDriver.getDriver().findElement(
+      By.xpath("//*[contains(@title,'" + sink + "')]")));
+    SeleniumHelper.dragAndDrop(
+      SeleniumDriver.getDriver().findElement(By.xpath("//*[contains(@class,'plugin-endpoint_" + source + "')]")),
+      SeleniumDriver.getDriver().findElement(By.xpath("//*[contains(@title,'" + sink + "')]")));
+  }
+
+  public static void selectBQ() throws InterruptedException {
+    SeleniumHelper.waitAndClick(CdfStudioLocators.sourceBigQuery);
+  }
+
+  public static void clickCloseButton() {
+    SeleniumHelper.waitAndClick(CdfStudioLocators.closeButton);
+  }
+
+  public static void sinkGcs() {
+    CdfStudioLocators.sink.click();
+    CdfStudioLocators.gcs.click();
   }
 }
--- a/src/main/java/io/cdap/e2e/pages/locators/CdfBigQueryPropertiesLocators.java
+++ b/src/main/java/io/cdap/e2e/pages/locators/CdfBigQueryPropertiesLocators.java
@@ -28,13 +28,13 @@ public class CdfBigQueryPropertiesLocators {
   @FindBy(how = How.XPATH, using = "//*[@data-cy='referenceName' and @class='MuiInputBase-input']")
   public static WebElement bigQueryReferenceName;
 
-  @FindBy(how = How.XPATH, using = "//*[@placeholder='Dataset the table belongs to']")
+  @FindBy(how = How.XPATH, using = "//*[@data-cy='dataset' and @class='MuiInputBase-input']")
   public static WebElement bigQueryDataSet;
 
   @FindBy(how = How.XPATH, using = "//*[@data-cy=\"project\" and @class='MuiInputBase-input']")
   public static WebElement projectID;
 
-  @FindBy(how = How.XPATH, using = "//*[@placeholder='Table to write to']")
+  @FindBy(how = How.XPATH, using = "//*[@data-cy='table' and @class='MuiInputBase-input']")
   public static WebElement bigQueryTable;
 
   @FindBy(how = How.XPATH, using = "//*[@data-cy='switch-truncateTable']")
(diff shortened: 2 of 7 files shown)


src/main/java/io/cdap/e2e/pages/locators/CdfBigQueryPropertiesLocators.java
```diff
@@ -60,4 +63,32 @@
 
   @FindBy(how = How.XPATH, using = "//*[@data-cy=\"cmekKey\" and @class='MuiInputBase-input']")
   public static WebElement cmekKey;
+
+  @FindBy(how = How.XPATH, using = "//*[@data-cy='partitionFrom' and @class='MuiInputBase-input']")
+  public static WebElement partitionStartDate;
```
tivv, line 74: This must not live here. The whole file should reside in google-cloud.

src/main/java/io/cdap/e2e/pages/actions/CdfStudioActions.java
```diff
@@ -81,4 +83,35 @@ public static String bannerErrorMessage() {
...
+
+  public static void  connectSourceAndSink(String source, String sink) {
+    SeleniumHelper.waitElementIsVisible(SeleniumDriver.getDriver().findElement(
+      By.xpath("//*[contains(@title,'" + sink + "')]")));
+    SeleniumHelper.dragAndDrop(
+      SeleniumDriver.getDriver().findElement(By.xpath("//*[contains(@class,'plugin-endpoint_" + source + "')]")),
+      SeleniumDriver.getDriver().findElement(By.xpath("//*[contains(@title,'" + sink + "')]")));
+  }
+
+  public static void selectBQ() throws InterruptedException {
```
tivv, line 104: BQ should not belong to e2e. Only basic GCS actions should be here that will be reused across different plugins

src/main/java/io/cdap/e2e/utils/CdfHelper.java
```diff
@@ -100,12 +101,12 @@ default void succeedPipelineVerification() {
...
   default int recordIn() {
     String incount = CdfPipelineRunLocators.recordsInCount.getText();
-    return Integer.parseInt(incount);
+    return Integer.valueOf(incount.replaceAll(",", StringUtils.EMPTY).toString());
   }
 
   default int recordOut() {
     String outcount = CdfPipelineRunLocators.recordsOutCount.getText();
-    return Integer.parseInt(outcount);
+    return Integer.valueOf(outcount.replaceAll(",", StringUtils.EMPTY).toString());
```
tivv: Why are you switching from `parseInt` to `valueOf`? Also I think `.toString()` is not needed

sawantpritam: Changed to parseInt and removed redundant .toString()

src/main/java/io/cdap/e2e/pages/actions/CdfStudioActions.java
sawantpritam: BigQuery is being used as Sink across different plugin testcases. thats why it is placed in e2e.
@ksahilCF

src/main/java/io/cdap/e2e/pages/locators/CdfBigQueryPropertiesLocators.java
sawantpritam: BigQuery is being used as Sink across different plugin testcases. thats why it is placed in e2e.
@ksahilCF

src/main/java/io/cdap/e2e/pages/actions/CdfStudioActions.java
tivv: Why? Do we have some test cases with non-google-cloud source and BQ sinks?

src/main/java/io/cdap/e2e/utils/GcpClient.java
```diff
@@ -63,6 +64,16 @@ private static int executeQuery(String query) throws InterruptedException, IOExc
     return 0;
   }
 
+  public static String executeSelectQuery(String query) throws InterruptedException, IOException {
```
itsankit-google: We already have an executeQuery() method here. Why can't we use it instead of adding another method?

sawantpritam: Existing method parses query result into integer.  Can not be used for String output.

itsankit-google: Well, then we should look at changing the existing method to return object and convert the results to the expected data type.

sawantpritam: We will need to find all the places where existing method is used and make change.
Instead can we mark existing method deprecated?

itsankit-google: Why do we need string output here? It looks like it returns the same (no of rows present in the table) instead of actual data. If they are same, wondering why do we need it as string instead of integer?

sawantpritam: No. We are using new method to execute queries which returns string(actual data from table cell).

src/main/java/io/cdap/e2e/pages/actions/CdfStudioActions.java
satyamsufi: Yes, we do have SAP data sources with BQ sink

tivv: OK

src/main/java/io/cdap/e2e/utils/GcpClient.java
tivv: It seems confusing. The only difference is data type, but the methods are `executeQuery` and `executeSelectQuery`. I agree that there should be one method that returns `Optional<String>` and any parsing can be done by caller or by a helper method with a clear name calling the main one

sawantpritam: Changed new executeSelectQuery() to return Optional<String>.
Calls to existing executeQuery() in e2e-tests project changed to new executeSelectQuery().
Existing method executeQuery() is public it is being used outside e2e-tests project as well. Marked it as Deprecated.

src/main/java/io/cdap/e2e/utils/GcpClient.java
```diff
@@ -63,6 +66,16 @@ public static int executeQuery(String query) throws InterruptedException, IOExce
     return 0;
   }
 
+  public static Optional<String> executeSelectQuery(String query) throws InterruptedException, IOException {
+    QueryJobConfiguration queryConfig = QueryJobConfiguration.newBuilder(query).build();
+    TableResult results = getBigQueryService().query(queryConfig);
+    String outputRowValue = StringUtils.EMPTY;
```
tivv: This won't work with Optionals. We don't want an Optional that contains ""

src/main/java/io/cdap/e2e/utils/GcpClient.java
```diff
@@ -42,17 +44,18 @@ public static int countBqQuery(String table) throws IOException, InterruptedExce
...
+    return executeSelectQuery(selectQuery).map(Integer::parseInt).orElse(0);
   }
 
   //Deleting the table
   public static void dropBqQuery(String table) throws IOException, InterruptedException {
     String projectId = SeleniumHelper.readParameters(ConstantsUtil.PROJECT_ID);
     String datasetName = SeleniumHelper.readParameters(ConstantsUtil.DATASET);
     String dropQuery = "DROP TABLE `" + projectId + "." + datasetName + "." + table + "`";
-    executeQuery(dropQuery);
+    executeSelectQuery(dropQuery);
```
tivv: It looks weird to call `executeSelectQuery` for drop. Please consider some other name

src/main/java/io/cdap/e2e/utils/GcpClient.java
```diff
@@ -42,17 +44,18 @@ public static int countBqQuery(String table) throws IOException, InterruptedExce
...
   }
 
+  @Deprecated
   public static int executeQuery(String query) throws InterruptedException, IOException {
     QueryJobConfiguration queryConfig = QueryJobConfiguration.newBuilder(query).build();
```
tivv, line 63: Delegate to `executeSelectQuery`

src/main/java/io/cdap/e2e/utils/GcpClient.java
sawantpritam: Changed.

src/main/java/io/cdap/e2e/utils/GcpClient.java
sawantpritam: changed name to executeQueryOnBQ

src/main/java/io/cdap/e2e/utils/GcpClient.java
sawantpritam: Done

src/main/java/io/cdap/e2e/utils/GcpClient.java
```diff
@@ -63,6 +70,16 @@ public static int executeQuery(String query) throws InterruptedException, IOExce
     return 0;
   }
 
+  public static Optional<String> executeQueryOnBQ(String query) throws InterruptedException, IOException {
+    QueryJobConfiguration queryConfig = QueryJobConfiguration.newBuilder(query).build();
+    TableResult results = getBigQueryService().query(queryConfig);
+    String outputRowValue = null;
+    if (results.getTotalRows() > 0) {
+      outputRowValue =  (String) results.getValues().iterator().next().get(0).getValue();
```
itsankit-google, line 77: One more question, as far as I can see this will return the first value of the first row of the resultSet if I am not wrong? Why do we need only that value? It earlier made sense because we were using above method to execute only count and drop query. So, I think it would be better to return the whole resultSet and the caller can fetch the required data from the result set! What do you think @tivv?

tivv: as far as I understand the whole goal is to get first item as this is the use case. May be that's also an answer to the method name, e.g. `getSoleQueryResult`

sawantpritam: Changed method name to `getSoleQueryResult`

tivv (approved):

PR #61, opened 2022-01-27
Selenium session teardown

sawantpritam:
@itsankit-google @rmstar  Closing selenium driver session after all scenarios. 
Cucumber does not have @BeforeAll and @AfterAll tags. This is workaround using Runtime.getRuntime().addShutdownHook in @Before.

### DIFF
--- a/src/main/java/stepsdesign/AfterActions.java
+++ b/src/main/java/stepsdesign/AfterActions.java
@@ -30,9 +30,13 @@ import org.openqa.selenium.WebDriver;
 public class AfterActions {
 
   @AfterStep
-  public static void tearDown(Scenario scenario) {
-    WebDriver driver = SeleniumDriver.getDriver();
-    byte[] screenshotBytes = ((TakesScreenshot) driver).getScreenshotAs(OutputType.BYTES);
-    scenario.embed(screenshotBytes, "image/png");
+  public static void takeScreenshot(Scenario scenario) {
+    try {
+      WebDriver driver = SeleniumDriver.getDriver();
+      byte[] screenshotBytes = ((TakesScreenshot) driver).getScreenshotAs(OutputType.BYTES);
+      scenario.embed(screenshotBytes, "image/png");
+    } catch (Exception e) {
+      scenario.write("Exception in capturing screenshot : " + e.getMessage() + " : " + e);
+    }
   }
 }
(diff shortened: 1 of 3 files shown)


src/main/java/stepsdesign/AfterActions.java
```diff
@@ -30,9 +30,13 @@
 public class AfterActions {
 
   @AfterStep
-  public static void tearDown(Scenario scenario) {
-    WebDriver driver = SeleniumDriver.getDriver();
-    byte[] screenshotBytes = ((TakesScreenshot) driver).getScreenshotAs(OutputType.BYTES);
-    scenario.embed(screenshotBytes, "image/png");
+  public static void takeScreenshot(Scenario scenario) {
```
rmstar, line 33: where is takeScreenShot called?

sawantpritam: It is not called explicitly anywhere. It gets executed after every cucumber test step with`@AfterStep` hook

src/main/java/stepsdesign/BeforeActions.java
```diff
@@ -32,11 +32,16 @@
 public class BeforeActions {
   public static Scenario scenario;
   public static File myObj;
+  public static boolean beforeAllFlag = true;
 
   @Before(order = 0)
   public void setUp(Scenario scenario) throws IOException {
+    if (beforeAllFlag) {
```
rmstar, line 39: why do we need this flag? Is setUp called multiple times?

rmstar: does it need to be volatile?

rmstar:
could we do something like @AfterStep(order=1000) to ensure that the cleanup step is executed at the end?

src/main/java/stepsdesign/BeforeActions.java
sawantpritam: Yes Cucumber `@Before` `setUp()` gets executed before each scenario.
Cucumber does not have `@BeforeAll `and `@Afterall` tags which will execute before first and after last scenario. We want to setup selenium driver before first scenario and teardown after final scenario. So flag is needed.

sawantpritam:
> could we do something like @afterstep(order=1000) to ensure that the cleanup step is executed at the end?

Order for @AfterStep is not required. The default value is 10000 if not mentioned explicitly.

rmstar (approved):

itsankit-google:
> could we do something like @afterstep(order=1000) to ensure that the cleanup step is executed at the end?

Its actually something good in cucumber that afterstep order is reverse, for eg, order=0 in after will be executed at the end.

PR #64, opened 2022-02-03
cdf common step definitions

sawantpritam:

### DIFF
--- a/src/main/java/stepsdesign/PipelineSteps.java
+++ b/src/main/java/stepsdesign/PipelineSteps.java
@@ -15,20 +15,30 @@
  */
 package stepsdesign;
 
-import io.cdap.e2e.pages.actions.CdfLogActions;
 import io.cdap.e2e.pages.actions.CdfPipelineRunAction;
 import io.cdap.e2e.pages.actions.CdfStudioActions;
+import io.cdap.e2e.pages.locators.CdfSchemaLocators;
 import io.cdap.e2e.pages.locators.CdfStudioLocators;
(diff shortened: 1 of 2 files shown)


src/main/java/stepsdesign/CdfCommonSteps.java
```diff
@@ -0,0 +1,126 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Cdf common stepDesigns.
+ */
+public class CdfCommonSteps implements CdfHelper {
```
itsankit-google: Let's move these steps also in pipeline steps class instead of new class.

src/main/java/stepsdesign/PipelineSteps.java
```diff
@@ -43,7 +42,7 @@
   @Then("Save the pipeline")
   public void saveThePipeline() {
     CdfStudioActions.pipelineName();
-    CdfStudioActions.pipelineNameIp("GCSTest" + UUID.randomUUID().toString());
+    CdfStudioActions.pipelineNameIp("TestPipeline" + UUID.randomUUID());
```
itsankit-google, line 117: Good catch. :)

src/main/java/stepsdesign/CdfCommonSteps.java
sawantpritam: Moved to `PipelineSteps`

itsankit-google (approved):

PR #65, opened 2022-02-07
Using projectID and dataset from plugin properties

sawantpritam:
@itsankit-google @rmstar  Please review.

### DIFF
--- a/src/main/resources/connectionParameters.properties
+++ b/src/main/resources/connectionParameters.properties
@@ -1,9 +1,5 @@
-# GCP project Id
-projectId=cdf-athena
 # Selenium browser window size
 windowSize=1920x1040
-# GCP datatest
-dataset=test_automation
 # CDAP application url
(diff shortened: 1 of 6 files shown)


src/main/resources/connectionParameters.properties
```diff
@@ -13,4 +9,8 @@ Project-ID=cdf-athena
 Data-Set=test_automation
 tableDemo=DemoCheck1
 delimiter@=@
+# GCP project Id
+projectId=cdf-athena
```
itsankit-google, line 13: why do we still need these properties here?

sawantpritam: We don't need this. Moved this to `# properties to be removed`.

I will remove this properties along with  all deprecated methods in new PR.

itsankit-google: Okay

itsankit-google (approved):

PR #68, opened 2022-02-15
logs exception handling and common step definition

sawantpritam:

### DIFF
--- a/src/main/java/io/cdap/e2e/utils/SeleniumHelper.java
+++ b/src/main/java/io/cdap/e2e/utils/SeleniumHelper.java
@@ -92,6 +92,14 @@ public class SeleniumHelper {
     act.dragAndDrop(from, to).build().perform();
   }
 
+  /**
+   * Performs click-and-hold at the location of the element, moves by a given offset, then releases the mouse.
+   */
+  public static void dragAndDropByOffset(WebElement element, int xOffset, int yOffset) {
+    Actions actions = new Actions(SeleniumDriver.getDriver());
+    actions.dragAndDropBy(element, xOffset, yOffset).build().perform();
+  }
+
   public static void clickObject(WebElement element) {
     element.click();
   }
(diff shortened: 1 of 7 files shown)


src/main/java/io/cdap/e2e/utils/SeleniumHelper.java
```diff
@@ -92,6 +92,11 @@ public static void dragAndDrop(WebElement from, WebElement to) {
     act.dragAndDrop(from, to).build().perform();
   }
 
+  public static void dragAndDropByOffset(WebElement element, int xOffset, int yOffset) {
```
itsankit-google, line 98: where is this method used? Can we have a javadoc for it?

src/main/java/stepsdesign/PipelineSteps.java
```diff
@@ -190,6 +217,20 @@ public void runThePipelineInRuntime() throws InterruptedException {
     CdfPipelineRunAction.runClick();
   }
 
+  @Then("Run the Pipeline in Runtime with runtime arguments")
+  public void runThePipelineInRuntimeWithRuntimeArguments() throws InterruptedException {
+    CdfPipelineRunAction.runClick();
+    for (Map.Entry<String, String> entry : runtimeArguments.entrySet()) {
+      if (SeleniumHelper.verifyElementPresent(
+        CdfStudioLocators.runtimeArgsValueLocator.replace("RUNTIME_ARGS_KEY", entry.getKey()))) {
```
itsankit-google, line 225: Iirc `runtime argument key` is automatically set after pipeline deployment, we need to only provide values. Why are we replacing keys as well?

sawantpritam: We are just checking if that `runtimetime argument key` is displayed and entering value in the text area next to that key.

src/main/java/io/cdap/e2e/utils/SeleniumHelper.java
sawantpritam: javadoc added.
When we have multiple sinks they gets overlapped on canvas so sometimes connection is established with wrong sink during automation. This method is used to align sinks properly on canvas to establish connection.

src/main/java/stepsdesign/PipelineSteps.java
itsankit-google: sg, thanks

src/main/java/io/cdap/e2e/utils/SeleniumHelper.java
itsankit-google: got it, thanks

itsankit-google (approved):
LGTM

PR #69, opened 2022-02-17
Divide SeleniumHelper class in relevant helper classes

poojantcs:
Changes: 

1. SeleniumHelper class has been divided into four classes:
WaitHelper, 
ElementHelper, 
PageHelper, 
AssertionHelper

2. A few common steps added in the PipelineSteps along with their Actions and Locators

@itsankit-google @rmstar @tivv Kindly review.
cc: @sawantpritam @ksahilCF

### DIFF
--- a/src/main/java/io/cdap/e2e/pages/actions/CdfStudioActions.java
+++ b/src/main/java/io/cdap/e2e/pages/actions/CdfStudioActions.java
@@ -17,28 +17,49 @@
 package io.cdap.e2e.pages.actions;
 
 import io.cdap.e2e.pages.locators.CdfStudioLocators;
-import io.cdap.e2e.utils.ConstantsUtil;
+import io.cdap.e2e.utils.ElementHelper;
 import io.cdap.e2e.utils.SeleniumDriver;
 import io.cdap.e2e.utils.SeleniumHelper;
+import io.cdap.e2e.utils.WaitHelper;
 import org.openqa.selenium.By;
-import org.openqa.selenium.JavascriptExecutor;
 import org.openqa.selenium.WebElement;
-
-import static io.cdap.e2e.utils.ConstantsUtil.JS_CLICK;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
- * Represents CdfStudioActions
+ * Represents Cdf Studio page Actions
  */
 public class CdfStudioActions {
+  private static final Logger logger = LoggerFactory.getLogger(CdfStudioActions.class);
   public static CdfStudioLocators cdfStudioLocators;
 
   static {
     cdfStudioLocators = SeleniumHelper.getPropertiesLocators(CdfStudioLocators.class);
   }
 
+  public static void selectDataPipelineType(String option) {
+    ElementHelper.clickOnElement(CdfStudioLocators.dataPipelineTypeDropdown);
+    /* TODO: Replace JsExecutor click with normal click: https://cdap.atlassian.net/browse/CDAP-18885 */
+    ElementHelper.clickOnElementUsingJsExecutor(CdfStudioLocators.locateDataPipelineTypeDropdownOption(option));
+  }
+
+  public static void expandPluginGroupIfNotAlreadyExpanded(String pluginGroupName) {
+    ElementHelper.clickIfDisplayed(CdfStudioLocators.locatePluginGroupCollapsed(pluginGroupName));
+    WaitHelper.waitForElementToBeDisplayed(CdfStudioLocators.locatePluginGroupExpanded(pluginGroupName));
+  }
+
+  public static void selectPluginFromList(String pluginName) {
+    logger.info("Click on plugin from the list: " + pluginName);
+    ElementHelper.clickOnElement(CdfStudioLocators.locatePluginNameInList(pluginName));
+  }
+
+  public static void navigateToPluginPropertiesPage(String pluginName) {
+    ElementHelper.hoverOverElement(CdfStudioLocators.locatePluginNodeInCanvas(pluginName));
+    ElementHelper.clickOnElement(CdfStudioLocators.locatePluginPropertiesButton(pluginName));
+  }
+
   /**
-   * @deprecated
-   * Use {@link io.cdap.e2e.utils.CdfHelper#selectSourcePlugin(String)}
+   * @deprecated Use {@link io.cdap.e2e.utils.CdfHelper#selectSourcePlugin(String)}
    */
   @Deprecated
   public static void selectGCS() throws InterruptedException {
@@ -47,8 +68,7 @@ public class CdfStudioActions {
   }
 
   /**
-   * @deprecated
-   * Use {@link io.cdap.e2e.utils.CdfHelper#selectSourcePlugin(String)}
+   * @deprecated Use {@link io.cdap.e2e.utils.CdfHelper#selectSourcePlugin(String)}
    */
   @Deprecated
   public static void clickSource() {
@@ -56,8 +76,8 @@ public class CdfStudioActions {
   }
 
   /**
-   * @deprecated
-   * Use {@link CdfStudioActions#clickSink()} and {@link io.cdap.e2e.utils.CdfHelper#selectSinkPlugin(String)}
+   * @deprecated Use {@link CdfStudioActions#clickSink()} and
+   * {@link io.cdap.e2e.utils.CdfHelper#selectSinkPlugin(String)}
    */
   @Deprecated
   public static void sinkBigQuery() {
@@ -87,8 +107,7 @@ public class CdfStudioActions {
   }
 
   public static void pipelineDeploy() {
-    JavascriptExecutor jse = (JavascriptExecutor) SeleniumDriver.getDriver();
-    jse.executeScript(JS_CLICK, CdfStudioLocators.pipelineDeploy);
+    ElementHelper.clickOnElement(CdfStudioLocators.pipelineDeploy);
   }
 
   public static String bannerErrorMessage() {
@@ -104,8 +123,7 @@ public class CdfStudioActions {
   }
 
   /**
-   * @deprecated
-   * Use either {@link io.cdap.e2e.utils.CdfHelper#openSinkPluginProperties(String)}
+   * @deprecated Use either {@link io.cdap.e2e.utils.CdfHelper#openSinkPluginProperties(String)}
    * or {@link io.cdap.e2e.utils.CdfHelper#openSourcePluginProperties(String)} as per plugin type.
    */
   @Deprecated
@@ -115,11 +133,10 @@ public class CdfStudioActions {
   }
 
   /**
-   * @deprecated
-   * Use {@link io.cdap.e2e.utils.CdfHelper#connectSourceAndSink(String, String)}
+   * @deprecated Use {@link io.cdap.e2e.utils.CdfHelper#connectSourceAndSink(String, String)}
    */
   @Deprecated
-  public static void  connectSourceAndSink(String source, String sink) {
+  public static void connectSourceAndSink(String source, String sink) {
     SeleniumHelper.waitElementIsVisible(SeleniumDriver.getDriver().findElement(
       By.xpath("//*[contains(@title,'" + sink + "')]")));
     SeleniumHelper.dragAndDrop(
@@ -128,8 +145,7 @@ public class CdfStudioActions {
   }
 
   /**
-   * @deprecated
-   * Use {@link io.cdap.e2e.utils.CdfHelper#selectSourcePlugin(String)}
+   * @deprecated Use {@link io.cdap.e2e.utils.CdfHelper#selectSourcePlugin(String)}
    */
   @Deprecated
   public static void selectBQ() throws InterruptedException {
@@ -137,12 +153,12 @@ public class CdfStudioActions {
   }
 
   public static void clickCloseButton() {
-    SeleniumHelper.waitAndClick(CdfStudioLocators.closeButton);
+    ElementHelper.clickOnElement(CdfStudioLocators.closeButton);
   }
 
   /**
-   * @deprecated
-   * Use {@link CdfStudioActions#clickSink()} and {@link io.cdap.e2e.utils.CdfHelper#selectSinkPlugin(String)}
+   * @deprecated Use {@link CdfStudioActions#clickSink()} and
+   * {@link io.cdap.e2e.utils.CdfHelper#selectSinkPlugin(String)}
    */
   @Deprecated
   public static void sinkGcs() {
--- a/src/main/java/io/cdap/e2e/pages/locators/CdfStudioLocators.java
+++ b/src/main/java/io/cdap/e2e/pages/locators/CdfStudioLocators.java
@@ -25,9 +25,12 @@ import org.openqa.selenium.support.How;
 import java.util.List;
 
 /**
- * Represents CdfStudioLocators
+ * Represents Cdf Studio Page Locators
  */
 public class CdfStudioLocators {
+  @FindBy(how = How.XPATH, using = "//div[contains(@class, 'left-top-section')]//select")
+  public static WebElement dataPipelineTypeDropdown;
+
   @FindBy(how = How.XPATH, using = "//*[@data-cy=\"plugin-GCSFile-batchsource\"]")
   public static WebElement source;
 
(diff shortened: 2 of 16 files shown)


src/main/java/io/cdap/e2e/utils/SeleniumHelper.java
```diff
@@ -77,6 +76,18 @@ public static void filemove(String name, String input) throws IOException, Inter
...
+  public static String readParameters(String property) throws IOException {
+    return connectProperties.getProperty(property);
+  }
+
+  public static <T> T getPropertiesLocators(Class<T> className) {
+    return PageFactory.initElements(SeleniumDriver.getDriver(), className);
+  }
+
+  /**
+   * @deprecated Use {@link ElementHelper#isElementDisplayed(WebElement)}
```
itsankit-google, line 88: Can we also call the new methods inside old deprecated methods to leverage the new changes??

src/main/java/io/cdap/e2e/utils/ElementHelper.java
```diff
@@ -0,0 +1,135 @@
...
+    actions.moveToElement(element).build().perform();
+  }
+
+  public static void clearElementValue(WebElement element) {
+    scrollToElement(element);
+    WaitHelper.waitForElementToBeDisplayed(element);
+    logger.info("Clear WebElement: " + element);
+    element.click();
+    element.clear();
+    while (!element.getAttribute("value").equals("")) {
```
itsankit-google: Replace while loop by for loop with some definite limit for eg, `1024` in `replaceElementValue`. We have seen cases in past where it gets stuck indefinitely due to some UI bug.

src/main/java/io/cdap/e2e/utils/ElementHelper.java
```diff
@@ -0,0 +1,135 @@
...
+    WaitHelper.waitForElementToBeDisplayed(source);
+    WaitHelper.waitForElementToBeDisplayed(target);
+    logger.info("Drag and drop between source: " + source + " and target: " + target);
+    actions.dragAndDrop(source, target).build().perform();
+  }
+
+  /**
+   * Performs click-and-hold at the location of the element, moves by a given offset, then releases the mouse.
+   */
+  public static void dragAndDropByOffset(WebElement element, int xOffset, int yOffset) {
```
itsankit-google, line 145: Why not waiting for element to be displayed in this method? Although seems unlikely to drag and drop an element which is not present but would be good to re-assure.

src/main/java/io/cdap/e2e/utils/ElementHelper.java
itsankit-google: Also please fail the test with exception and include nesessary debug information in the exception. e.g. If you expect some value to become empty and it's not after some definite attempts, include the final value in the exception.

src/main/java/io/cdap/e2e/utils/ElementHelper.java
poojantcs: updated

src/main/java/io/cdap/e2e/utils/SeleniumHelper.java
poojantcs: updated.

src/main/java/io/cdap/e2e/utils/ElementHelper.java
poojantcs: updated.

src/main/java/io/cdap/e2e/utils/ConstantsUtil.java
```diff
@@ -31,13 +31,17 @@
...
   public static final String PASS_CASE = "First case passed";
   public static final String ERROR_MSG_COLOR = "#a40403";
   public static final String ERROR_MSG_MANDATORY = "Required property 'PROPERTY' has no value.";
   public static final String DEFAULT_PLUGIN_PROPERTIES_FILE = "pluginParameters.properties";
   public static final String DEFAULT_ERROR_PROPERTIES_FILE = "errorMessage.properties";
   public static final String VALIDATION_SUCCESS_MESSAGE = "validationSuccessMessage";
   public static final String VALIDATION_ERROR_MESSAGE = "validationErrorMessage";
+  public static final int DEFAULT_TIMEOUT = 60;
+  public static final int PAGE_LOAD_TIMEOUT = 10;
+  public static final int IMPLICIT_TIMEOUT = 5;
```
itsankit-google: Please add unit to timeouts. For eg, `DEFAULT_TIMEOUT_SECONDS`. Also, can we add a small comment with each of them when and where they are used?

src/main/java/io/cdap/e2e/utils/ElementHelper.java
```diff
@@ -0,0 +1,152 @@
...
+        break;
+      }
+
+      element.sendKeys(Keys.BACK_SPACE);
+    }
+
+    if (!flag) {
+      Assert.fail("Element's current value is not cleared - " + element + " - "
+        + "Current value :" + element.getAttribute("value"));
+    }
```
itsankit-google: Since `replaceElementValue` is also using the same logic to clear the element value. Better to combine it within one common method and call it from both.

itsankit-google: Or may be call `clearElementValue` from `replaceElementValue` to first clear the element.

src/main/java/io/cdap/e2e/utils/ElementHelper.java
```diff
@@ -0,0 +1,152 @@
...
+  private static final JavascriptExecutor js = (JavascriptExecutor) SeleniumDriver.getDriver();
+  private static final Actions actions = new Actions(SeleniumDriver.getDriver());
+
+  public static void scrollToElement(WebElement element) {
+    logger.info("Scrolling to the element: " + element);
+    js.executeScript("arguments[0].scrollIntoView(true);", element);
+  }
+
+  public static void clickOnElement(WebElement element) {
+    scrollToElement(element);
```
itsankit-google: Does it not need to check for visibility before scrolling to that element??

poojantcs: Yes, to click or basically to interact with an element, it has to be in the visible state. Otherwise [ElementNotInteractableException](https://www.selenium.dev/selenium/docs/api/java/org/openqa/selenium/ElementNotInteractableException.html) gets thrown.

src/main/java/io/cdap/e2e/utils/ConstantsUtil.java
poojantcs: Comments added.

src/main/java/io/cdap/e2e/utils/ElementHelper.java
poojantcs: Updated.

src/main/java/io/cdap/e2e/pages/actions/CdfStudioActions.java
```diff
@@ -18,27 +18,49 @@
...
     cdfStudioLocators = SeleniumHelper.getPropertiesLocators(CdfStudioLocators.class);
   }
 
+  public static void selectDataPipelineTypeAsBatch() {
+    ElementHelper.clickOnElement(CdfStudioLocators.dataPipelineTypeDropdown);
+    ElementHelper.clickOnElementUsingJsExecutor(
+      CdfStudioLocators.getDataPipelineTypeDropdownOption(ConstantsUtil.BATCH));
+  }
+
+  public static void selectDataPipelineTypeAsRealtime() {
```
itsankit-google: I think we can combine `selectDataPipelineTypeAsBatch` and `selectDataPipelineTypeAsRealtime` in one method. Something like this `selectDataPipelineType(String type)`

src/main/java/io/cdap/e2e/pages/locators/CdfStudioLocators.java
```diff
@@ -145,4 +148,25 @@ public static WebElement macroButton(String pluginProperty) {
...
+    String xpath = "//div[contains(@class, 'left-top-section')]//select//option[contains(@label, '" + option + "')]";
+    return SeleniumDriver.getDriver().findElement(By.xpath(xpath));
+  }
+
+  public static WebElement getPluginNameInList(String pluginName) {
+    String xpath = "//div[contains(@class, 'plugin-name')][normalize-space(text()) = '" + pluginName + "']";
+    return SeleniumDriver.getDriver().findElement(By.xpath(xpath));
+  }
+
+  public static WebElement getPluginNameInDiagramContainer(String pluginName) {
```
itsankit-google: Where is this used? I think more intuitive name is required here. What is diagram container here? Do you mean pipeline studio canvas? If yes, we can rename it to something like `locatePluginNodeInCanvas(String pluginName)`.

itsankit-google: Similarly in other method names replace `get` with `locate`

src/main/java/io/cdap/e2e/pages/actions/CdfStudioActions.java
poojantcs: Updated.

src/main/java/io/cdap/e2e/pages/locators/CdfStudioLocators.java
poojantcs: Yes, it's for the plugin node in the studio canvas. Updated the locator names as suggested.

src/main/java/io/cdap/e2e/utils/CdfHelper.java
```diff
@@ -46,19 +45,16 @@
...
     SeleniumDriver.openPage(SeleniumHelper.readParameters(ConstantsUtil.CDFURL));
-    try {
-      SeleniumDriver.getDriver().switchTo().alert().accept();
-      SeleniumDriver.waitForPageToLoad();
-    } catch (NoAlertPresentException Ex) {
-      SeleniumDriver.waitForPageToLoad();
-    }
+    PageHelper.acceptAlertIfPresent();
+    /* TODO: Remove static wait once https://cdap.atlassian.net/browse/CDAP-18862 is fixed */
+    WaitHelper.waitForPageToLoad(ConstantsUtil.SMALL_TIMEOUT_SECONDS);
```
itsankit-google: Looks like this needs to be increased otherwise tests are failing, currently Studio Page is rendered two times as described by the JIRA.

poojantcs: I have increased it to 5 secs. Let's see how the execution goes.

src/main/java/io/cdap/e2e/pages/actions/CdfStudioActions.java
```diff
@@ -17,28 +17,48 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(CdfStudioActions.class);
   public static CdfStudioLocators cdfStudioLocators;
 
   static {
     cdfStudioLocators = SeleniumHelper.getPropertiesLocators(CdfStudioLocators.class);
   }
 
+  public static void selectDataPipelineType(String option) {
+    ElementHelper.clickOnElement(CdfStudioLocators.dataPipelineTypeDropdown);
+    ElementHelper.clickOnElementUsingJsExecutor(CdfStudioLocators.locateDataPipelineTypeDropdownOption(option));
```
itsankit-google, line 43: Please open a JIRA for it as discussed offline and add it to the code as a comment. 
Context - using JsExecutor since alert is not getting closed. Also add the screenshot in the JIRA.

src/main/java/io/cdap/e2e/utils/ElementHelper.java
```diff
@@ -0,0 +1,159 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(ElementHelper.class);
+  private static final JavascriptExecutor js = (JavascriptExecutor) SeleniumDriver.getDriver();
+  private static final Actions actions = new Actions(SeleniumDriver.getDriver());
+
+  public static void scrollToElement(WebElement element) {
+    logger.info("Scrolling to the element: " + element);
+    actions.moveToElement(element).perform();
+  }
+
+  public static void scrollToElementUsingJsExecutor(WebElement element) {
```
itsankit-google, line 46: Please add a javadoc to jsExecutor methods explaning these methods should only be used when necessary since test should reproduce what a user would do with the browser.

itsankit-google (commented):
Overall looks good, add javadoc to methods for better readability.

src/main/java/io/cdap/e2e/pages/actions/CdfStudioActions.java
poojantcs: Ticket: https://cdap.atlassian.net/browse/CDAP-18885

src/main/java/io/cdap/e2e/utils/ElementHelper.java
poojantcs: Added.

src/main/java/io/cdap/e2e/utils/ElementHelper.java
```diff
@@ -0,0 +1,169 @@
...
+    if (isElementDisplayed(element, timeoutInSeconds)) {
+      clickOnElement(element, timeoutInSeconds);
+    }
+  }
+
+  public static void clickIfDisplayed(WebElement element) {
+    clickIfDisplayed(element, ConstantsUtil.SMALL_TIMEOUT_SECONDS);
+  }
+
+  public static void sendKeys(WebElement element, String keys) {
```
rmstar, line 85: please add javadoc for all public methods

src/main/java/io/cdap/e2e/utils/ElementHelper.java
```diff
@@ -0,0 +1,169 @@
...
+    WaitHelper.waitForElementToBeDisplayed(element);
+    scrollToElement(element);
+    logger.info("Clearing WebElement: " + element);
+    boolean flag = false;
+    element.click();
+    element.clear();
+
+    for (int i = 0; i < limit; i++) {
+      if (element.getAttribute("value").equals("")) {
+        flag = true;
```
rmstar: nit: no need for a flag here, just return if element.getAttribute("value").equals(""). Assert.fail if we come out of the for loop

rmstar (commented):
Looks good, please add javadoc for all public methods

src/main/java/io/cdap/e2e/pages/locators/CdfStudioLocators.java
```diff
@@ -25,9 +25,12 @@
 import java.util.List;
 
 /**
- * Represents CdfStudioLocators
+ * Represents Cdf Studio page Locators
```
itsankit-google: nit - Page

src/main/java/io/cdap/e2e/utils/ElementHelper.java
poojantcs: Sure, they will be added as part of https://cdap.atlassian.net/browse/CDAP-18884 in a separate PR.

src/main/java/io/cdap/e2e/utils/ElementHelper.java
poojantcs: Updated

src/main/java/io/cdap/e2e/pages/locators/CdfStudioLocators.java
poojantcs: Updated.

itsankit-google (approved):
LGTM

src/main/java/io/cdap/e2e/utils/WaitHelper.java
```diff
@@ -0,0 +1,95 @@
...
+      }
+    }
+  }
+
+  public static void waitForPageToLoad(long staticWaitInSeconds) {
+    waitForPageToLoad();
+
+    try {
+      logger.info("Additional static wait for page to load and stabilize: " + staticWaitInSeconds);
+      Thread.sleep(staticWaitInSeconds);
```
rmstar: throw an exception in waitForPageToLoad() if it doesn't load within the timeout. In general, avoid waiting forever as it causes tests to hang, makes it hard to debug - throw an exception if the condition you're waiting for is not satisfied within some reasonable timeout.

rmstar (commented):
Just one more comment.

src/main/java/io/cdap/e2e/utils/WaitHelper.java
poojantcs: Added timeout for the page load condition.

rmstar (approved):

PR #75, opened 2022-03-04
GCS and BQ locators and actions change

sawantpritam:

1. Cleaned up xpaths of existing locators to remove class.
2. Added locators and actions.

### DIFF
--- a/src/main/java/io/cdap/e2e/pages/actions/CdfBigQueryPropertiesActions.java
+++ b/src/main/java/io/cdap/e2e/pages/actions/CdfBigQueryPropertiesActions.java
@@ -17,48 +17,22 @@
 package io.cdap.e2e.pages.actions;
 
 import io.cdap.e2e.pages.locators.CdfBigQueryPropertiesLocators;
-import io.cdap.e2e.pages.locators.CdfStudioLocators;
 import io.cdap.e2e.utils.ElementHelper;
-import io.cdap.e2e.utils.PluginPropertyUtils;
 import io.cdap.e2e.utils.SeleniumHelper;
 
 import java.io.IOException;
-
-import static io.cdap.e2e.utils.ConstantsUtil.AUTOMATION_TEST;
-import static io.cdap.e2e.utils.ConstantsUtil.DATASET;
-import static io.cdap.e2e.utils.ConstantsUtil.ONE;
-import static io.cdap.e2e.utils.ConstantsUtil.PROJECT_ID;
 
 /**
  * Represents CdfBigQueryPropertiesActions
  */
 public class CdfBigQueryPropertiesActions {
-  public static CdfBigQueryPropertiesLocators cdfBigQueryPropertiesLocators;
 
   static {
-    cdfBigQueryPropertiesLocators = SeleniumHelper.getPropertiesLocators(CdfBigQueryPropertiesLocators.class);
-  }
-
-  /**
-   * @deprecated Call individual actions as per test scenario in step design file.
-   */
-  @Deprecated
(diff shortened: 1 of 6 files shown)


src/main/java/io/cdap/e2e/pages/actions/CdfBigQueryPropertiesActions.java
```diff
@@ -133,4 +129,47 @@ public static void enterTemporaryBucketName(String bucket) throws IOException {
   public static void clickPreviewData() {
     SeleniumHelper.waitAndClick(CdfBigQueryPropertiesLocators.previewData);
   }
+
+  public static void enterPartitionField(String partitionField) {
+    CdfBigQueryPropertiesLocators.partitionField.sendKeys(partitionField);
```
itsankit-google: can we use `ElementHelper` in all the actions?

itsankit-google: This applies to all GCS actions as well.

sawantpritam: Element helper has additional checks and steps which are not required in all actions.
Element helper should be called only when necessary.

itsankit-google: Why we don't want these additional checks here? From my understanding, it checks for presence of element and scrolls to the element correctly which makes the test more stable, isn't it??

itsankit-google: @poojantcs WDYT?

poojantcs: Hi @itsankit-google , we should use the ElementHelper as much as we can. For instances where the ElementHelper is not working, we should look to either enhance the existing functions or add new ones to it.

sawantpritam: Updated to use ElementHelper.

src/main/java/io/cdap/e2e/pages/locators/CdfGCSLocators.java
```diff
@@ -124,4 +107,19 @@
...
+
+  public static WebElement pathFilenameOnly(String value) {
+    return SeleniumDriver.getDriver().findElement(By.xpath("//input[@name='filenameOnly' and @value='" + value + "']"));
+  }
+
+  public static WebElement recursive(String value) {
+    return SeleniumDriver.getDriver().findElement(By.xpath("//input[@name='recursive' and @value='" + value + "']"));
+  }
+
+  public static WebElement getListItem(String option) {
```
itsankit-google: Is it a general locator or selecting item from particular list?? From locator name it is not clear. If its generic it should be in some general PluginPropertiesLocators.

sawantpritam: This is currently used only in GCSActions but can be used across plugins to get dropdown list item.
I will rename it and add TODO to move it to CdfPluginPropertiesLocators.java as this file is not yet merged.

itsankit-google (approved):
LGTM, Please squash commits.

src/main/java/io/cdap/e2e/pages/locators/CdfGCSLocators.java
sawantpritam: Done.

sawantpritam:
Squashed commits.

PR #79, opened 2022-03-23
Add common plugin steps

poojantcs:
* Steps and actions for clicking macro button of a Property and setting the value in it
* Steps and actions for retrieving the number of 'records out' and 'records in' on the pipeline run page
* Common steps added for plugin properties page

### DIFF
--- a/src/main/java/io/cdap/e2e/utils/ConstantsUtil.java
+++ b/src/main/java/io/cdap/e2e/utils/ConstantsUtil.java
@@ -22,6 +22,7 @@ public class ConstantsUtil {
   public static final String AUTOMATION_TEST = "automation_test";
   public static final String PROJECT_ID = "projectId";
   public static final String DATASET = "dataset";
+  public static final String TABLE = "tableName";
   public static final String SAMPLE_SIZE = "1000";
   public static final String DELIMITER = "delimiter@";
   public static final String VALUE = "value";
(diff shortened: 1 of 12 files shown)


src/main/java/io/cdap/e2e/utils/ConstantsUtil.java
```diff
@@ -42,6 +43,7 @@
   public static final String ERROR_MSG_MANDATORY = "Required property 'PROPERTY' has no value.";
   public static final String DEFAULT_PLUGIN_PROPERTIES_FILE = "pluginParameters.properties";
   public static final String DEFAULT_ERROR_PROPERTIES_FILE = "errorMessage.properties";
+  public static final String DEFAULT_DATACY_ATTRIBUTES_FILE = "pluginDataCyAttributes.properties";
```
itsankit-google, line 46: Why do we need this file?

poojantcs: With Common steps introduced in this PR, we will be passing the data-cy attribute of a property while configuring the plugin. Same data-cy attribute will be passed in all the tests. We thought of adding this property file with the data-cy mapping so we will only need to make a change in one place in case there is a change in the data-cy attribute of a property.

itsankit-google (approved):
lgtm

PR #84, opened 2022-04-28
Preview button action and locator changes

sawantpritam:
@itsankit-google Please review.

### DIFF
--- a/src/main/java/io/cdap/e2e/pages/actions/CdfStudioActions.java
+++ b/src/main/java/io/cdap/e2e/pages/actions/CdfStudioActions.java
@@ -171,10 +171,10 @@ public class CdfStudioActions {
   }
 
   /**
-   * Click on the Preview menu button to Open/Close the Preview menu
+   * Click on the Preview menu button to Close the Preview menu
    */
-  public static void previewSelect() {
-    ElementHelper.clickOnElement(CdfStudioLocators.preview);
+  public static void closePreviewMenu() {
+    ElementHelper.clickOnElement(CdfStudioLocators.previewActiveButton);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/main/java/io/cdap/e2e/pages/actions/CdfStudioActions.java
```diff
@@ -171,10 +171,10 @@ public static void openPipelinePreviewMenuAndClickOnRunButton() {
   }
 
   /**
-   * Click on the Preview menu button to Open/Close the Preview menu
+   * Click on the Preview menu button to Close the Preview menu
```
itsankit-google, line 174: On a closer look, the `Preview` Widget can be used for both Open/Close operations. I understand, currently we might be using it for close but still better to keep it naming generic so that can be reused if we need to perform Open operation as well.

itsankit-google (changes requested):
fixing the locator is fine, but we don't need to change the description and naming. Thanks!

src/main/java/io/cdap/e2e/pages/actions/CdfStudioActions.java
sawantpritam: We have separate action specific to open preview menu. 
As locator is different when preview button is active vs inactive.
https://github.com/cdapio/cdap-e2e-tests/blob/43b0d2424fb413642c88e4ada74b74bbb445f2a4/src/main/java/io/cdap/e2e/pages/actions/CdfStudioActions.java#L154

itsankit-google: I see, then it makes sense.

itsankit-google (approved):
LGTM

PR #85, opened 2022-04-28
Add pipeline wait steps

poojantcs:


src/main/java/io/cdap/e2e/pages/actions/CdfPipelineRunAction.java
```diff
@@ -93,6 +95,13 @@ public static Boolean isProvisioning() {
     return validatePipelineExpectedStatus(CdfPipelineRunLocators.provisioningStatus);
   }
 
+  /**
+   * Wait till the Pipeline's status changes to Running
+   */
+  public static void waitForPipelineToBeInRunningStatus() {
```
itsankit-google: Does it make sense to make this method generic and accept `status` and `timeout` as params?
It takes some time for pipeline to transition in `running` state, so it would be better to keep `timeout` configurable.

src/main/java/io/cdap/e2e/pages/actions/CdfPipelineRunAction.java
```diff
@@ -119,6 +128,14 @@ public static void waitTillPipelineRunCompletes(long timeoutInSeconds) {
     ));
   }
 
+  /**
+   * Stop the Pipeline
+   */
+  public static void stopPipeline() {
+    ElementHelper.clickOnElement(CdfPipelineRunLocators.stop);
+    WaitHelper.waitForElementToBeDisplayed(CdfPipelineRunLocators.stoppedStatus);
```
itsankit-google: Similarly here, transition to `stopped` state will not be instantaneous always. It would be better to have a separate step after clicking on stop to wait for transition to `stopped` state.

src/main/java/io/cdap/e2e/pages/actions/CdfPipelineRunAction.java
```diff
@@ -23,10 +23,12 @@
 import io.cdap.e2e.utils.PageHelper;
 import io.cdap.e2e.utils.SeleniumDriver;
 import io.cdap.e2e.utils.SeleniumHelper;
+import io.cdap.e2e.utils.WaitHelper;
 import org.junit.Assert;
 import org.openqa.selenium.StaleElementReferenceException;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.support.ui.ExpectedConditions;
+import org.openqa.selenium.support.ui.Wait;
```
itsankit-google: unused import?

src/main/java/io/cdap/e2e/pages/actions/CdfPipelineRunAction.java
poojantcs: Updated the method to expect Status and Timeout.

src/main/java/io/cdap/e2e/pages/actions/CdfPipelineRunAction.java
poojantcs: This wait step will wait till the Pipeline is transitioned to stopped state with the default timeout (60 s). Also added an overloaded method which expects a Timeout.

src/main/java/io/cdap/e2e/pages/actions/CdfPipelineRunAction.java
poojantcs: Yes, left it by mistake. Imports are optimized now.

src/main/java/io/cdap/e2e/pages/locators/CdfPipelineRunLocators.java
```diff
@@ -74,8 +74,8 @@ public static WebElement locatePipelineStatus(String status) {
   @FindBy(how = How.XPATH, using = "(//*[contains(@class, 'metric-records-out-label')])[1]/following-sibling::span")
   public static WebElement recordsInCount;
 
-  @FindBy(how = How.XPATH, using = "(//*[contains(@class, 'metric-records-out-label')])[2]" +
-    "/following-sibling::span[text()!='0']")
+  @FindBy(how = How.XPATH, using = "//*[@data-type='batchsink']//*[contains(text(),'In')]" +
```
itsankit-google: The plugin might not be `batchsink` always. Can we use some better locator or may be make `batch-sink` configurable?

src/main/java/io/cdap/e2e/pages/actions/CdfPipelineRunAction.java
itsankit-google: I still feel it is not a good idea to wait here, instead after clicking stop pipeline, use this step to wait till pipeline is in running state, https://github.com/cdapio/cdap-e2e-tests/blob/develop/src/main/java/stepsdesign/PipelineSteps.java#L353

itsankit-google: And then use this step to verify that pipeline status is `stopped`.
https://github.com/cdapio/cdap-e2e-tests/blob/develop/src/main/java/stepsdesign/PipelineSteps.java#L358

poojantcs: Removed the line waiting for the status to transition to Stopped.

src/main/java/io/cdap/e2e/pages/locators/CdfPipelineRunLocators.java
poojantcs: Updated the locator and made it configurable. It will expect the plugin node name.

itsankit-google (approved):
LGTM. Please squash commits.

poojantcs:
> LGTM. Please squash commits.

Commits are squashed.

PR #87, opened 2022-05-03
CDAP-19179:Add login logic for HDF in e2e test

jockech:
After enabling the security in ODF, we need the login logic to reach the studio page.

Tested and verified locally.

### DIFF
--- a/src/main/java/io/cdap/e2e/pages/actions/HdfSignInActions.java
+++ b/src/main/java/io/cdap/e2e/pages/actions/HdfSignInActions.java
@@ -16,27 +16,34 @@
 
 package io.cdap.e2e.pages.actions;
 
-import io.cdap.e2e.pages.locators.OdfSignInLocator;
+import io.cdap.e2e.pages.locators.HdfSignInLocator;
+import io.cdap.e2e.utils.ElementHelper;
 import io.cdap.e2e.utils.SeleniumHelper;
 
 import java.io.IOException;
 
-import static io.cdap.e2e.utils.ConstantsUtil.ODFPAZWRD;
-import static io.cdap.e2e.utils.ConstantsUtil.ODFUSERNAME;
+import static io.cdap.e2e.utils.ConstantsUtil.HDFPAZWRD;
+import static io.cdap.e2e.utils.ConstantsUtil.HDFUSERNAME;
 
 /**
- * Represents OdfSignInActions
+ * Represents HdfSignInActions
  */
-public class OdfSignInActions {
-  public static OdfSignInLocator odfSignInLocator;
+public class HdfSignInActions {
+  public static HdfSignInLocator hdfSignInLocator;
+  private static boolean logged = false;
 
   static {
-    odfSignInLocator = SeleniumHelper.getPropertiesLocators(OdfSignInLocator.class);
+    hdfSignInLocator = SeleniumHelper.getPropertiesLocators(HdfSignInLocator.class);
+  }
+
+  public static boolean logged() {
+    return logged;
   }
 
   public static void login() throws InterruptedException, IOException {
-    odfSignInLocator.odfUsername.sendKeys(SeleniumHelper.readParameters(ODFUSERNAME));
-    odfSignInLocator.odfPassword.sendKeys(SeleniumHelper.readParameters(ODFPAZWRD));
-    odfSignInLocator.loginButton.click();
+    ElementHelper.sendKeys(hdfSignInLocator.hdfUsername, SeleniumHelper.readParameters(HDFUSERNAME));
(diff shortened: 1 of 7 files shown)


src/main/java/io/cdap/e2e/utils/CdfHelper.java
```diff
@@ -47,6 +48,9 @@ default void openCdf() throws IOException, InterruptedException {
     SeleniumDriver.openPage(SeleniumHelper.readParameters(ConstantsUtil.CDFURL));
     PageHelper.acceptAlertIfPresent();
     WaitHelper.waitForPageToLoad();
+    OdfSignInActions.login();
```
itsankit-google: I don't think it is a good idea to hook odf specific step here, since this is used in all tests which runs on cdap-sandbox/cdf.

@rmstar WDYT?

jockech: I agree. Is there anyway to differentiate the ODF and CDF for this case like adding a flag?

rmstar: yes this will break tests running on sandbox. Can we put this in the place where other ODF specific test code lives?

itsankit-google: Is this a one time operation? Because afaik, this method is called by every scenario, what happens if the user is already logged in?

src/main/java/io/cdap/e2e/pages/actions/OdfSignInActions.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+public class OdfSignInActions {
+  public static OdfSignInLocator odfSignInLocator;
+
+  static {
+    odfSignInLocator = SeleniumHelper.getPropertiesLocators(OdfSignInLocator.class);
+  }
+
+  public static void login() throws InterruptedException, IOException {
+    odfSignInLocator.odfUsername.sendKeys(SeleniumHelper.readParameters(ODFUSERNAME));
```
itsankit-google: Please use relevant `ElementHelper` methods.
https://github.com/cdapio/cdap-e2e-tests/blob/develop/src/main/java/io/cdap/e2e/utils/ElementHelper.java#L131

src/main/java/io/cdap/e2e/pages/actions/OdfSignInActions.java
```diff
@@ -0,0 +1,42 @@
...
+    odfSignInLocator.odfPassword.sendKeys(SeleniumHelper.readParameters(ODFPAZWRD));
+    odfSignInLocator.loginButton.click();
```
itsankit-google: Similarly here, https://github.com/cdapio/cdap-e2e-tests/blob/develop/src/main/java/io/cdap/e2e/utils/ElementHelper.java#L75

src/main/java/io/cdap/e2e/utils/CdfHelper.java
itsankit-google: May be we can set an environment variable as a flag to trigger this action. But we will need to ensure it does not break when user is already logged in.

rmstar:
We should probably rename Odf -> Hdf in the PR

jockech:
> We should probably rename Odf -> Hdf in the PR

Done

src/main/java/io/cdap/e2e/utils/CdfHelper.java
jockech: I have added the environment variable which can be modified before the test.

To fix the logic for login, I have added a static variable to specify if we need to handle the login after the first successful login. Let me know if there is any concern of concurrent issue.

Have tested the bigquery source testcase and got the expected result.

src/main/java/io/cdap/e2e/pages/actions/HdfSignInActions.java
```diff
@@ -0,0 +1,49 @@
...
+
+  public static boolean logged() {
+    return logged;
+  }
+
+  public static void login() throws InterruptedException, IOException {
+    ElementHelper.sendKeys(hdfSignInLocator.hdfUsername, SeleniumHelper.readParameters(HDFUSERNAME));
+    ElementHelper.sendKeys(hdfSignInLocator.hdfPassword, SeleniumHelper.readParameters(HDFPAZWRD));
+    ElementHelper.clickOnElement(hdfSignInLocator.loginButton);
+    logged = true;
```
itsankit-google: I think this is okay. On a second thought, is it better to check for `NoSuchElementException` on the first locator and assume that the user is already logged in?

itsankit-google (approved):
LGTM

PR #89, opened 2022-05-06
Enhance Schema validation and Logs related methods

poojantcs:
Two Pipeline steps are added:
1. To verify the schema rows of expanded hierarchical fields (record type and array type) separately
2. To verify list of log entries in the Advanced logs page

### DIFF
--- a/src/main/java/io/cdap/e2e/pages/actions/CdfPluginPropertiesActions.java
+++ b/src/main/java/io/cdap/e2e/pages/actions/CdfPluginPropertiesActions.java
@@ -28,6 +28,7 @@ import io.cdap.e2e.utils.SeleniumDriver;
 import io.cdap.e2e.utils.SeleniumHelper;
 import io.cdap.e2e.utils.WaitHelper;
 import org.junit.Assert;
+import org.openqa.selenium.By;
 import org.openqa.selenium.WebElement;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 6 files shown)


src/main/java/io/cdap/e2e/pages/actions/CdfPluginPropertiesActions.java
```diff
@@ -290,6 +291,36 @@ public static Map<String, String> getOutputSchema() {
...
+   * @param fieldType
+   * @return
+   */
+  public static Boolean scrollDownInOutputSchemaUntilFieldEntry(String fieldName, String fieldType) {
+    By locator = CdfPluginPropertiesLocators.locatorOfOutputSchemaFieldTypeEntry(fieldName, fieldType);
+    boolean isFieldTypeEntryPresent = WaitHelper.waitForElementToBeOptionallyPresent(locator, 1);
+    int totalSchemaRowsInit = CdfPluginPropertiesLocators.outputSchemaRows.size();
+    int totalSchemaRowsUpdated;
+
+    while (!isFieldTypeEntryPresent) {
```
itsankit-google, line 311: why do we need this extra for loop?
When is it possible that we need to traverse the whole output schema?

poojantcs: We came across a few use cases where the list of Output Schema fields is longer than 50. 
The current application behaviour is such that it only shows the first ~20 fields and only shows the rest of the fields once we start scrolling down. The fields at the bottom are not even present in the DOM. 

Hence, we are checking for the presence of the field and if it is not available then only scrolling down and checking for its presence again.

itsankit-google: I see, thanks for the info. Could you please add a comment in the code block?

poojantcs: Added the comment. Also squashed the commits since it was only an addition of a comment.

itsankit-google (approved):
lgtm

PR #90, opened 2022-05-12
Connection management & common verification methods

sawantpritam:

### DIFF
--- a/src/main/java/io/cdap/e2e/pages/actions/CdfConnectionActions.java
+++ b/src/main/java/io/cdap/e2e/pages/actions/CdfConnectionActions.java
@@ -1,0 +1,221 @@
+/*
+ * Copyright © 2022 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.e2e.pages.actions;
+
+import io.cdap.e2e.pages.locators.CdfConnectionLocators;
+import io.cdap.e2e.pages.locators.CdfPluginPropertiesLocators;
+import io.cdap.e2e.utils.AssertionHelper;
+import io.cdap.e2e.utils.ConstantsUtil;
+import io.cdap.e2e.utils.ElementHelper;
+import io.cdap.e2e.utils.PageHelper;
+import io.cdap.e2e.utils.PluginPropertyUtils;
+import io.cdap.e2e.utils.SeleniumDriver;
+import io.cdap.e2e.utils.SeleniumHelper;
+import io.cdap.e2e.utils.WaitHelper;
+import org.openqa.selenium.StaleElementReferenceException;
+import org.openqa.selenium.WebElement;
+
+import java.io.IOException;
+
+/**
+ * Represents Cdf connection page and properties Actions
+ */
+public class CdfConnectionActions {
+
+  static {
+    SeleniumHelper.getPropertiesLocators(CdfPluginPropertiesLocators.class);
+    SeleniumHelper.getPropertiesLocators(CdfConnectionLocators.class);
+  }
+
+  /**
+   * Open Wrangler connections page
+   */
+  public static void openWranglerConnectionsPage() throws IOException {
+    SeleniumDriver.openPage(SeleniumHelper.readParameters(ConstantsUtil.WRANGLER_CONNECTIONS_URL));
+    PageHelper.acceptAlertIfPresent();
+    WaitHelper.waitForPageToLoad();
+  }
+
+  /**
+   * Click on the Browser connections button inside Plugin's properties page
+   */
+  public static void clickBrowseConnectionsButton() {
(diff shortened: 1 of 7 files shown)


src/main/java/io/cdap/e2e/pages/actions/CdfConnectionActions.java
```diff
@@ -0,0 +1,204 @@
...
+    if (connectionNameFromPluginPropertiesFile == null) {
+      WaitHelper.waitForElementToBeDisplayed(CdfPluginPropertiesLocators.locateElementWithText(connectionName));
+      return;
+    }
+    WaitHelper.waitForElementToBeDisplayed(CdfPluginPropertiesLocators
+                                             .locateElementWithText(connectionNameFromPluginPropertiesFile));
+  }
+
+  /**
+   * Select data row for connection
```
itsankit-google, line 109: What do you mean by `data row`, could you please mention some examples in comments?

src/main/java/io/cdap/e2e/pages/actions/CdfConnectionActions.java
```diff
@@ -0,0 +1,204 @@
...
+        if (attempts == 4) {
+          throw e;
+        }
+      }
+      attempts++;
+    }
+  }
+
+  /**
+   * Select data for connection using name
```
itsankit-google, line 136: Similarly here, what is `data` here?

src/main/java/io/cdap/e2e/pages/actions/CdfConnectionActions.java
```diff
@@ -0,0 +1,204 @@
...
+   *
+   * @param dataName If dataName is present in {@link ConstantsUtil#DEFAULT_PLUGIN_PROPERTIES_FILE} as a key
```
itsankit-google, line 146: Similarly, what is `dataName` here?

src/main/java/io/cdap/e2e/pages/actions/CdfConnectionActions.java
```diff
@@ -0,0 +1,204 @@
...
+      actualDataPath = dataPath;
+    }
+    String[] dataPathSplit = actualDataPath.split("/");
+    String title = dataPathSplit[dataPathSplit.length - 1];
+    WaitHelper.waitForElementToBePresent(CdfConnectionLocators.locatorOfDataTableTitle(title));
+    WaitHelper.waitForElementToBeDisplayed(CdfConnectionLocators.dataTable);
+  }
+
+  /**
+   * Click on Create Pipeline button in Wrangler datatable page
```
itsankit-google: nit `Wrangler datatable page` -> `Wrangler Workspace page`

src/main/java/io/cdap/e2e/pages/actions/CdfConnectionActions.java
```diff
@@ -0,0 +1,204 @@
...
+   */
+  public static void clickCreatePipelineButton() {
+    ElementHelper.clickOnElement(CdfConnectionLocators.createPipelineButton);
+  }
+
+  /**
+   * Select the type of pipeline to create
+   *
+   * @param pipelineType If pipelineType is present in {@link ConstantsUtil#DEFAULT_PLUGIN_PROPERTIES_FILE} as key
```
itsankit-google, line 208: I think priority should be `inverse` everywhere, if the property is passed as a param select it, otherwise from `Properties File` and else `fail`.

src/main/java/io/cdap/e2e/pages/actions/CdfConnectionActions.java
sawantpritam: Added example in comments.

src/main/java/io/cdap/e2e/pages/actions/CdfConnectionActions.java
sawantpritam: Added example in comments.

src/main/java/io/cdap/e2e/pages/actions/CdfConnectionActions.java
sawantpritam: Added example in comments.

src/main/java/io/cdap/e2e/pages/actions/CdfConnectionActions.java
sawantpritam: Updated.

src/main/java/io/cdap/e2e/pages/actions/CdfConnectionActions.java
sawantpritam: Priority is correct.
As we are keeping actual values in the properties file. 

Example - 
`Then Enter input plugin property: "datasetProjectId" with value: "projectId"`

In this case if we `inverse` priority. It will enter value as "projectId" in input instead of actual value "cdf-athena" from properties file.

itsankit-google: I see, we are passing plugin_prop as the param.

src/main/java/io/cdap/e2e/pages/actions/CdfConnectionActions.java
```diff
@@ -0,0 +1,217 @@
...
+  public static void selectConnectionDataWithName(String dataName) {
+    String dataToSelect = PluginPropertyUtils.pluginProp(dataName);
+    if (dataToSelect == null) {
+      dataToSelect = dataName;
+    }
+    selectConnectionDataRow(dataToSelect);
+  }
+
+  /**
+   * Select data rows for connection traversing through directories using provided path
```
itsankit-google, line 159: Please add an example `path` in comments.

sawantpritam: Added example path in comment.

itsankit-google (approved):

PR #93, opened 2022-05-24
Connection management - common locators and actions

sawantpritam:
@itsankit-google Please review.

### DIFF
--- a/src/main/java/io/cdap/e2e/pages/actions/CdfConnectionActions.java
+++ b/src/main/java/io/cdap/e2e/pages/actions/CdfConnectionActions.java
@@ -218,4 +218,51 @@ public class CdfConnectionActions {
       CdfConnectionLocators.locatorOfPipelineTypeButton(pipelineTypeToSelect)));
   }
 
+  /**
+   * Expand the connections for provided connection type
+   *
+   * @param connectionType i.e. BigQuery / GCS / File
+   */
+  public static void expandConnections(String connectionType) {
+    String isExpanded = ElementHelper.getElementAttribute(CdfConnectionLocators.locateConnectionType(connectionType),
+                                                          "aria-expanded");
+    if (isExpanded.equalsIgnoreCase("false")) {
+      ElementHelper.clickOnElement(CdfConnectionLocators.locateConnectionType(connectionType));
+    }
+  }
+
+  /**
+   * Open the connection's action menu
+   *
+   * @param connectionType i.e. BigQuery / GCS / File
+   * @param connectionName If connectionName is present in {@link ConstantsUtil#DEFAULT_PLUGIN_PROPERTIES_FILE} as key
+   *                       then actual connectionName is fetched from it
+   *                       else connectionName param used as it is.
+   */
+  public static void openConnectionActionMenu(String connectionType, String connectionName) {
+    String actualConnectionName = PluginPropertyUtils.pluginProp(connectionName);
+    if (actualConnectionName == null) {
+      actualConnectionName = connectionName;
+    }
+    ElementHelper.clickOnElement(
+      CdfConnectionLocators.locateConnectionActionMenu(connectionType, actualConnectionName));
+  }
+
+  /**
+   * Select the action for connection
+   *
+   * @param connectionType i.e. BigQuery / GCS / File
+   * @param connectionName If connectionName is present in {@link ConstantsUtil#DEFAULT_PLUGIN_PROPERTIES_FILE} as key
+   *                       then actual connectionName is fetched from it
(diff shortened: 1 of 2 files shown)


src/main/java/io/cdap/e2e/pages/locators/CdfConnectionLocators.java
```diff
@@ -50,4 +51,20 @@ public static By locatorOfDataTableTitle(String title) {
...
   }
+
+  public static WebElement locateConnectionType(String connectionType) {
+    return SeleniumDriver.getDriver().findElement(
+      By.xpath("//div[@data-cy='categorized-connection-type-" + connectionType + "']"));
+  }
+
+  public static WebElement locateConnectionActionMenu(String connectionName) {
+    return SeleniumDriver.getDriver().findElement(
+      By.xpath("//*[text()='" + connectionName + "']/parent::a/following-sibling::div"));
```
itsankit-google: can this locator be improved so that it is more specific? Currently, according to my understanding, it will select the `div` following `connectionName` on the dom.

src/main/java/io/cdap/e2e/pages/actions/CdfConnectionActions.java
```diff
@@ -218,4 +218,43 @@ public static void selectPipelineType(String pipelineType) {
       CdfConnectionLocators.locatorOfPipelineTypeButton(pipelineTypeToSelect)));
   }
 
+  /**
+   * Expand the connections for provided connection type
+   *
+   * @param connectionType i.e. BigQuery / GCS / File
+   */
+  public static void expandConnections(String connectionType) {
+    ElementHelper.clickOnElement(CdfConnectionLocators.locateConnectionType(connectionType));
```
itsankit-google: If the connection list is expanded by default, for eg, in GCS (in case of CDF), we might end up collapsing it by clicking it, can we add a check to ensure we only click if it is not already expanded?

src/main/java/io/cdap/e2e/pages/locators/CdfConnectionLocators.java
sawantpritam: Changed locator

src/main/java/io/cdap/e2e/pages/actions/CdfConnectionActions.java
sawantpritam: Added check.

src/main/java/io/cdap/e2e/pages/locators/CdfConnectionLocators.java
```diff
@@ -50,4 +51,20 @@ public static By locatorOfDataTableTitle(String title) {
...
     return By.xpath("//a//*[text()='" + pipelineType + "']");
   }
+
+  public static WebElement locateConnectionType(String connectionType) {
+    return SeleniumDriver.getDriver().findElement(
+      By.xpath("//div[@data-cy='categorized-connection-type-" + connectionType + "']"));
+  }
+
+  public static WebElement locateConnectionActionMenu(String connectionName) {
+    return SeleniumDriver.getDriver().findElement(By.xpath("//*[text()='" + connectionName + "']" +
```
itsankit-google: Can we use `span[@data-cy='connection-connectionType-connectionName']` instead of `text()`?

src/main/java/io/cdap/e2e/pages/locators/CdfConnectionLocators.java
```diff
@@ -50,4 +51,20 @@ public static By locatorOfDataTableTitle(String title) {
...
+                 "/parent::a/following-sibling::div[contains(@class,'actions-popover')]"));
+  }
+
+  public static WebElement locateConnectionAction(String connectionName, String action) {
+    return SeleniumDriver.getDriver().findElement(
+      By.xpath("//*[text()='" + connectionName + "']/parent::a/following-sibling::div//li[text()='" + action + "']"));
```
itsankit-google: similarly here.

src/main/java/io/cdap/e2e/pages/locators/CdfConnectionLocators.java
sawantpritam: Updated locator with data-cy

src/main/java/io/cdap/e2e/pages/locators/CdfConnectionLocators.java
sawantpritam: Updated locator with data-cy

itsankit-google (approved):
Lgtm

PR #100, opened 2022-06-21
click on the element and move to avoid overlap

jockech:
Right now, selecting the same plugin will cause overlap issue.
e.g:
image
Then it's impossible to enter the `File` plugin properties.

Adding `moveElement` logic will help to resolve the issue.

### DIFF
--- a/src/main/java/io/cdap/e2e/pages/actions/CdfStudioActions.java
+++ b/src/main/java/io/cdap/e2e/pages/actions/CdfStudioActions.java
@@ -452,4 +452,14 @@ public class CdfStudioActions {
   public static void closeStatusBanner() {
     ElementHelper.clickOnElement(CdfStudioLocators.statusBannerCloseButton);
   }
+
+  /**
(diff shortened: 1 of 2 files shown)


itsankit-google (commented):
I think we already have the actions for this functionality https://github.com/cdapio/cdap-e2e-tests/blob/23c0640d96573ccfb05da238333a50cc16ef4ba3/src/main/java/io/cdap/e2e/utils/ElementHelper.java#L251
Please correct me if I am wrong.

jockech:
> I think we already have the actions for this functionality
> 
> https://github.com/cdapio/cdap-e2e-tests/blob/23c0640d96573ccfb05da238333a50cc16ef4ba3/src/main/java/io/cdap/e2e/utils/ElementHelper.java#L251
> 
> 
> Please correct me if I am wrong.

yes. thanks for pointing out.

itsankit-google (approved):
lgtm :)

PR #110, opened 2022-07-05
Update run_e2e_test_module.py

ksahilCF:
@itsankit-google @rmstar please review

### DIFF
--- a/src/main/scripts/run_e2e_test_module.py
+++ b/src/main/scripts/run_e2e_test_module.py
@@ -39,10 +39,9 @@ if __name__ == "__main__":
     os.system("export JAVA_OPTS=-Xmx24G")
     run_shell_command(f"sandbox/{sandbox_dir}/bin/cdap sandbox start")
    
-
+    
     # Build the plugin
     os.chdir("plugin")
-
     module_to_build : str = sys.argv[1]
     print(f"Building plugin {module_to_build}")
     run_shell_command(f"mvn clean package -pl {module_to_build} -am -DskipTests")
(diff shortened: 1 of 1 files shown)


src/main/scripts/run_e2e_test_module.py
```diff
@@ -38,11 +38,14 @@ def run_shell_command(cmd):
...
     os.system("export JAVA_OPTS=-Xmx24G")
     run_shell_command(f"sandbox/{sandbox_dir}/bin/cdap sandbox start")
-   
-
+    
+    #Building Framework
+    os.chdir("e2e")
+    print("Preparing e2e framework")
+    run_shell_command("mvn clean install")
+
```
itsankit-google: Why are we building the framework before installing the plugin?

src/main/scripts/run_e2e_test_module.py
```diff
@@ -70,19 +73,18 @@ def run_shell_command(cmd):
...
-    run_shell_command("mvn clean install")
+    os.chdir("../..")
+    print("cwd:", os.getcwd())
+    print("ls:", os.listdir())
     print("Running e2e integration tests")
-    assertion_error = None
-    os.chdir(f"../plugin/{module_to_build}")
     try:
-        run_shell_command("mvn clean install -P e2e-tests")
+        run_shell_command(f"mvn install -pl {module_to_build} -amd -P e2e-tests")
```
itsankit-google: We already built the module in Line#51 ig, why do we need to do `mvn clean install` again??

src/main/scripts/run_e2e_test_module.py
ksahilCF: changed the sequence to plugin build first

src/main/scripts/run_e2e_test_module.py
ksahilCF: mvn verify is used to run test now

src/main/scripts/run_e2e_test_module.py
```diff
@@ -70,19 +69,23 @@ def run_shell_command(cmd):
     assert res.ok or print(res.text)
     res=requests.put(f"http://localhost:11015/v3/namespaces/default/artifacts/{plugin_name}/versions/{plugin_version}/properties", json=plugin_properties)
     assert res.ok or print(res.text)
-    # Run e2e tests
+    
+     # Run e2e tests
+    print("cwd:", os.getcwd())
+    print("ls:", os.listdir())
```
itsankit-google: Are these debug statements which were pushed by mistake?

itsankit-google (commented):
Please attach an example workflow in description in which this script is tested.

src/main/scripts/run_e2e_test_module.py
```diff
@@ -39,10 +39,9 @@ def run_shell_command(cmd):
     os.system("export JAVA_OPTS=-Xmx24G")
     run_shell_command(f"sandbox/{sandbox_dir}/bin/cdap sandbox start")
    
-
+    
     # Build the plugin
     os.chdir("plugin")
-
     module_to_build : str = sys.argv[1]
```
itsankit-google, line 45: Can we use optional named args instead of positional args?

src/main/scripts/run_e2e_test_module.py
ksahilCF: removed

src/main/scripts/run_e2e_test_module.py
ksahilCF: I will implement it in separate PR

ksahilCF:
@itsankit-google -  Please find the workflow https://github.com/data-integrations/sap-plugins/actions/runs/2621330291

itsankit-google (approved):
lgtm. Pls squash commits.

PR #113, opened 2022-07-07
CDAP-19392 & CDAP-19390

sawantpritam:
CDAP-19392 - It is Failing while waiting for visibility of plugin properties button and not for plugin node.
Properties button is displayed on hover of the plugin node.
When Pipeline is created from wrangler connection, Plugin node is displayed in left corner first and then aligned in the middle on canvas. Hover action happens before plugin get aligned, so properties button gets hidden in meantime.

### DIFF
--- a/src/main/java/io/cdap/e2e/pages/actions/CdfStudioActions.java
+++ b/src/main/java/io/cdap/e2e/pages/actions/CdfStudioActions.java
@@ -25,6 +25,7 @@ import io.cdap.e2e.utils.SeleniumDriver;
 import io.cdap.e2e.utils.SeleniumHelper;
 import io.cdap.e2e.utils.WaitHelper;
 import org.openqa.selenium.By;
+import org.openqa.selenium.TimeoutException;
 import org.openqa.selenium.WebElement;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


src/main/java/io/cdap/e2e/pages/actions/CdfStudioActions.java
```diff
@@ -111,7 +112,20 @@ public static void verifyPluginNodeIsDisplayedOnCanvas(String pluginName, long t
    * @param pluginName
    */
   public static void navigateToPluginPropertiesPage(String pluginName) {
-    ElementHelper.hoverOverElement(CdfStudioLocators.locatePluginNodeInCanvas(pluginName));
+    int attempts = 0;
+    while (attempts < 5) {
```
rmstar: please add a comment explaining why this is necessary (basically what you have added in the PR description)

rmstar: nit: can we use a constant for max attempts?

rmstar (approved):

src/main/java/io/cdap/e2e/pages/actions/CdfStudioActions.java
sawantpritam: 1. Comment added explaining need. 
2. Used constant for max attempts.

itsankit-google (approved):

PR #116, opened 2022-07-12
add generic helper methods

mrahanjam:
This PR adds two generic helper methods that we came across during writing the E2E tests for tethering. I figured these could be used in future in other tests as well.


itsankit-google (approved):
lgtm

PR #117, opened 2022-07-20
Add Snapshot Release Workflow

PranavNair0001:


PranavNair0001:
Similar successful workflow run result: https://github.com/PranavNair0001/google-cloud/actions/runs/2702332435

.github/workflows/snapshot-release.yml
```diff
@@ -0,0 +1,72 @@
...
+            CDAP_OSSRH_USERNAME:cdapio-github-builds/CDAP_OSSRH_USERNAME
+            CDAP_OSSRH_PASSWORD:cdapio-github-builds/CDAP_OSSRH_PASSWORD
+            CDAP_GPG_PASSPHRASE:cdapio-github-builds/CDAP_GPG_PASSPHRASE
+
+      - name: Checkout Repository
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ inputs.ref }}
+
+      - name: Set up JDK 1.8
```
tivv: Why do we need this? Runner should already have java

.github/workflows/snapshot-release.yml
```diff
@@ -0,0 +1,72 @@
...
+      - name: Set up Apache Maven Central
+        uses: actions/setup-java@v1
+        with: # running setup-java again overwrites the settings.xml
+          java-version: 1.8
+          server-id: ossrh
+          server-username: CDAP_OSSRH_USERNAME
+          server-password: CDAP_OSSRH_PASSWORD
+          gpg-passphrase: CDAP_GPG_PASSPHRASE
+
+      - name: Publish to Apache Maven Central
```
tivv: I don't think it's maven central, is it?

.github/workflows/snapshot-release.yml
```diff
@@ -0,0 +1,72 @@
...
+        if: ${{ endsWith(steps.version.outputs.version, 'SNAPSHOT') }}
+        run: mvn -DskipTests clean deploy -P release-snapshot -Dgpg.passphrase=$CDAP_GPG_PASSPHRASE
```
tivv: Do we need to pass a parameter? I though it will pick it up from settings.xml

.github/workflows/snapshot-release.yml
itsankit-google: +1, remove redundant steps.

.github/workflows/e2e.yml
```diff
@@ -1,4 +1,4 @@
-# Copyright © 2021 Cask Data, Inc.
+# Copyright © 2022 Cask Data, Inc.
```
rmstar: nit: no need to update year when you update a file

.github/workflows/snapshot-release.yml
PranavNair0001: Setup Java action makes it possible to configure and add a server profile in `settings.xml` for Maven Release. Hence the requirement, it's not due to necessity of Java.
Much recently, many GitHub workflows for Maven release are following this method, since its an easy automated setup of `settings.xml`. Found in several articles too.

.github/workflows/snapshot-release.yml
PranavNair0001: I'll update the name. Good catch. Thank you

.github/workflows/snapshot-release.yml
PranavNair0001: I tried, it doesn't work. Since our `maven-gpg-plugin` in `pom.xml` is configured for automated signing without human/user intervention in CLI

.github/workflows/e2e.yml
PranavNair0001: Sure

.github/workflows/snapshot-release.yml
PranavNair0001: I have removed the redundant usage

itsankit-google (approved):
lgtm

PR #133, opened 2022-07-29
Plugin property common action for key value pairs

sawantpritam:
This PR contains :
1. Common action to enter key value pairs.

image

2. Common action to change datatype of fields in output schema

image

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -86,7 +86,7 @@ jobs:
           sha: ${{ github.event.workflow_run.head_sha }}
 
   run-snapshot-release:
-    if: ${{ github.event.workflow_run.event != 'pull_request' && endsWith(needs.build.outputs.version, 'SNAPSHOT') }}
+    if: ${{ github.event.workflow_run.event == 'push' && endsWith(needs.build.outputs.version, 'SNAPSHOT') }}
     needs: build
     uses: ./.github/workflows/snapshot-release.yml
     with:


src/main/java/io/cdap/e2e/pages/actions/CdfPluginPropertiesActions.java
```diff
@@ -852,4 +869,33 @@ public static void verifyRadioButtonPluginPropertySelectedValue(String pluginPro
...
+    }
+    Map<String, String> properties =
+      JsonUtils.convertKeyValueJsonArrayToMap(PluginPropertyUtils.pluginProp(jsonKeyValuePairs));
+    int index = 0;
+    for (Map.Entry<String, String> entry : properties.entrySet()) {
+      ElementHelper.sendKeys(CdfPluginPropertiesLocators.locatePropertyKey(
+        pluginPropertyDataCyAttribute, index), entry.getKey());
+      ElementHelper.sendKeys(CdfPluginPropertiesLocators.locatePropertyValue(
+        pluginPropertyDataCyAttribute, index), entry.getValue());
+      ElementHelper.clickOnElement(CdfPluginPropertiesLocators.locatePropertyAddRowButton(
```
itsankit-google: I think this will end up adding an extra row than required?

sawantpritam: Updated logic to add row.

src/main/java/io/cdap/e2e/pages/actions/CdfPluginPropertiesActions.java
```diff
@@ -113,6 +113,23 @@ public static void selectMacroActionOfOutputSchemaProperty() {
...
+   *                              is the json array with field name and datatype
+   *                              ({@link ConstantsUtil#DEFAULT_PLUGIN_PROPERTIES_FILE})
+   */
+  public static void changeDataTypesInOutputSchema(String jsonFieldList) {
+    Map<String, String> outputSchemaFieldList =
+      JsonUtils.convertKeyValueJsonArrayToMap(PluginPropertyUtils.pluginProp(jsonFieldList));
+    for (Map.Entry<String, String> entry : outputSchemaFieldList.entrySet()) {
+      ElementHelper.clickOnElement(CdfSchemaLocators.outputSchemaDataTypeDropdown(entry.getKey()));
+      SeleniumDriver.getDriver()
+        .findElement(CdfSchemaLocators.outputSchemaDataTypeOption(entry.getKey(), entry.getValue())).click();
```
itsankit-google: Can we use `ElementHelper` here?

sawantpritam: No. Element helper click method first waits for element to be clickable, then scroll to that element and then click.
Output schema datatype options list element is not scrollable. We get exception while scrolling -
`element not interactable: [object HTMLOptionElement] has no size and location`
So direct click action is used.

PR #134, opened 2022-08-01
Use cdap-e2e-framework snapshot from maven repo in plugin repos

sawantpritam:

### DIFF
--- a/src/main/scripts/run_e2e_test.py
+++ b/src/main/scripts/run_e2e_test.py
@@ -34,6 +34,7 @@ def run_shell_command(cmd):
 parser=argparse.ArgumentParser()
 parser.add_argument('--testRunner', help='TestRunner class to execute tests')
 parser.add_argument('--module', help='Module for which tests need to be run')
+parser.add_argument('--framework', help='Pass this param if workflow is triggered from e2e framework repo')
 args=parser.parse_args()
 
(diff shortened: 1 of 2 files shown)


src/main/scripts/run_e2e_test.py
```diff
@@ -109,11 +110,16 @@ def run_shell_command(cmd):
...
-os.chdir("e2e")
-run_shell_command("mvn clean install")
+if args.framework:
+    print("Preparing e2e framework")
+    os.chdir("e2e")
+    run_shell_command("mvn clean install")
+    os.chdir("../plugin")
+else:
+    os.chdir("plugin")
+    run_shell_command("mvn dependency:purge-local-repository -DmanualInclude=io.cdap.tests.e2e:cdap-e2e-framework")
```
itsankit-google, line 120: Why do we need a purge? We are always running the tests on fresh runner, so should not need it.

sawantpritam: We are caching `.m2 repo` in workflow. 
https://github.com/cdapio/cdap-e2e-tests/blob/5095fd778008227704ea392da4aa9fa606f31845/.github/workflows/e2e.yml#L46

itsankit-google (approved):

PR #136, opened 2022-08-11
Database driver installation

sawantpritam:

Workflow run : https://github.com/data-integrations/database-plugins/actions/runs/2818569612

### DIFF
--- a/src/main/scripts/run_e2e_test.py
+++ b/src/main/scripts/run_e2e_test.py
@@ -22,6 +22,7 @@ import zipfile
 import shutil
 import sys
 import argparse
+import urllib.request
 
 def run_shell_command(cmd):
     process = subprocess.run(cmd.split(" "), stderr=subprocess.PIPE)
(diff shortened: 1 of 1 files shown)


src/main/scripts/run_e2e_test.py
```diff
@@ -102,9 +102,46 @@ def run_shell_command(cmd):
...
+        driver_prop = '[{"name":"mysql","type":"jdbc","className":"com.mysql.jdbc.Driver","description":"Plugin for the MySQL JDBC driver"}]'
+        artifact_name = "mysql-connector-java"
+        artifact_version = "8.0.25"
+        install_driver = "true"
+    elif module_to_build == 'oracle-plugin':
+        jar_file_name = "ojdbc7.jar"
+        driver_prop = '[{"name":"oracle","type":"jdbc","className":"oracle.jdbc.driver.OracleDriver","description":"Plugin for the Oracle 12c JDBC driver"}]'
+        artifact_name = "ojdbc7"
+        artifact_version = "7"
+        install_driver = "true"
```
itsankit-google: Instead of using dumping everything in the same script, can we have a separate `jdbc-drivers.yml` file which contains all the props like -

```
modules:
   module_to_build: `postgress-plugin`
      jar: postgresql-9.4.1211.jar
      driver_prop: .....
 ```

File: src/main/scripts/run_e2e_test.py [hunk 2]
```diff
@@ -103,8 +104,22 @@ def run_shell_command(cmd):
 
 if module_to_build:
     os.chdir("../../..")
+    if module_to_build == 'postgresql-plugin' or module_to_build == 'mssql-plugin' or module_to_build == 'mysql-plugin' or module_to_build == 'oracle-plugin':
```
itsankit-google: Can we get the list of modules as well from yaml? 
something like `if module_to_build in driver_details['modules']:`

src/main/scripts/run_e2e_test.py
sawantpritam: Done.

src/main/scripts/run_e2e_test.py
sawantpritam: Done.

PR #137, opened 2022-08-12
use the new github-actions cdap-build link

itsankit-google:


src/main/scripts/run_e2e_test.py
```diff
@@ -39,7 +39,7 @@ def run_shell_command(cmd):
 
 # Start CDAP sandbox
 print("Downloading CDAP sandbox")
-sandbox_url = "https://builds.cask.co/artifact/CDAP-BUT/shared/build-latestSuccessful/SDK/cdap/cdap-standalone/target/cdap-sandbox-6.8.0-SNAPSHOT.zip"
+sandbox_url = "https://github.com/cdapio/cdap-build/releases/download/latest/cdap-sandbox-6.8.0-SNAPSHOT.zip"
```
sechegaray, line 42: Should we use latest, being that it's develop?

itsankit-google: Release name is `latest` but we will still need to specify the zip file name, it is the way how cdap builds the standalone zip file name. I don't know if it would be good to override that by unzipping & zipping it again or renaming the zip file.

sechegaray (approved):

PR #140, opened 2022-08-18
locator changes as per latest sandbox UI

sawantpritam:

### DIFF
--- a/src/main/java/io/cdap/e2e/pages/locators/CdfStudioLocators.java
+++ b/src/main/java/io/cdap/e2e/pages/locators/CdfStudioLocators.java
@@ -39,7 +39,7 @@ public class CdfStudioLocators {
   @FindBy(how = How.XPATH, using = "//*[@data-cy='plugin-BigQueryTable-batchsink']")
   public static WebElement bigQueryObject;
 
-  @FindBy(how = How.XPATH, using = "//div[@data-cy='plugin-Sink-group']//span[normalize-space(text())='Sink']")
+  @FindBy(how = How.XPATH, using = "//div[@data-cy='plugin-Sink-group']//*[normalize-space(text())='Sink']")
   public static WebElement sink;
 
   @FindBy(how = How.XPATH, using = "//*[contains(@class,'plugin-endpoint_SAP-ODP')]")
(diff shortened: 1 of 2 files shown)


src/main/java/io/cdap/e2e/utils/ElementHelper.java
```diff
@@ -73,7 +73,7 @@ public static void clickOnElementUsingJsExecutor(WebElement element) {
    */
   public static void clickOnElement(WebElement element) {
     WaitHelper.waitForElementToBeClickable(element);
-    scrollToElement(element);
+    scrollToElementUsingJsExecutor(element);
```
itsankit-google: Why is this change needed? 
Using js executor does not simulate user actions.

sawantpritam: Normal `scrollToElement(element)` is not working in new left side panel.

image

itsankit-google: This change might be related to the recent migration of Pipeline Studio from angular to react.
But we cannot make this action use JS Executor since this is being used across all tests.
Please check with @mrahanjam & @sumengwang.

sumengwang: We want to fix the ui to match the e2e tests as much as possible. It will be good to just identify the issues first. But I believe this pr is necessary.

sumengwang (dismissed):
LGTM

src/main/java/io/cdap/e2e/utils/ElementHelper.java
itsankit-google: Can we please open a JIRA to track this? I am fine with this change if it is temporary but in longer term using JS Executor might not be the correct way to interact with UI elements in end-to-end tests.

sumengwang: https://github.com/cdapio/cdap-ui/pull/605/files
this should fix the scroll issue. The footer will no longer block the elements. no need for jsexecutor

sawantpritam: Removed use of JS Executor. Above change to fix scroll issue is not available in latest sandbox successful build. @itsankit-google Please re-trigger e2e tests after next successful sandbox build.

itsankit-google (commented):
Please look into test failures.

itsankit-google (approved):

PR #144, opened 2022-08-19
Added script to install multiple dependencies from submodules and then run E2E tests from a submodule

fernst:
This can be useful in multi-module projects if structured like:

```
Parent project
├ Submodule 1
├ Submodule 2
├ Submodule 3
└ E2E tests submodule
```

Note that the parent project and all submodules must be build before using this script.

Example usage in https://github.com/data-integrations/google-cloud/blob/gcp-modules-v3/.github/workflows/e2e.yml#L71

### DIFF
--- a/src/main/scripts/run_e2e_test_multi_module.py
+++ b/src/main/scripts/run_e2e_test_multi_module.py
@@ -1,0 +1,134 @@
+# Copyright © 2022 Cask Data, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not
+# use this file except in compliance with the License. You may obtain a copy of
+# the License at
+
+# http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations under
+# the License.
+
+import io
+import json
+import os
+import requests
+import subprocess
+import xml.etree.ElementTree as ET
+import zipfile
+import shutil
+import sys
+import argparse
+import urllib.request
+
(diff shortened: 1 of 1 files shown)


src/main/scripts/run_e2e_test_multi_module.py
```diff
@@ -0,0 +1,134 @@
...
+for dependency in args.dependency:
+    print(f"Installing dependency: {dependency}")
+    os.chdir(dependency)
+
+    # Get plugin artifact name and version from pom.xml.
+    root = ET.parse('pom.xml').getroot()
+    plugin_name = root.find('{http://maven.apache.org/POM/4.0.0}artifactId').text
+    plugin_version = root.find('{http://maven.apache.org/POM/4.0.0}version').text
+
+    os.chdir('target')
```
itsankit-google, line 78: We will need to build the `dependency module` first then only we can install it from the `target` folder right??

fernst: yes, the entire project needs to be built using `mvn package` before installing all dependencies and running E2E tests

itsankit-google: Looks like this script isn't taking care of that.

The older script first builds the plugin module and then installs it.
ref - https://github.com/cdapio/cdap-e2e-tests/blob/develop/src/main/scripts/run_e2e_test.py#L69

fernst: Should this be the responsibility of the plugin? With multi-module projects there may be optimizations (such as only building certain modules) or using multiple threads when compiling. I wanted to leave this for the parent project to decide how to handle.

itsankit-google: Discussed offline to keep it as the responsibility of caller that module is built before running the script.
Can we add a comment here in the script for future reference??

itsankit-google (approved):
lgtm

PR #152, opened 2022-10-07
open sink only if needed

mrahanjam:
This PR makes a small change in the openSink method. I noticed that if the test is selecting multiple sinks at a time, every other click results in shrinking the sink group.

### DIFF
--- a/src/main/java/io/cdap/e2e/pages/actions/CdfStudioActions.java
+++ b/src/main/java/io/cdap/e2e/pages/actions/CdfStudioActions.java
@@ -309,7 +309,10 @@ public class CdfStudioActions {
    * Click on the Plugin group title: Sink
    */
   public static void clickSink() {
-    ElementHelper.clickOnElement(CdfStudioLocators.sink);
+    // Check if Sink is not already expanded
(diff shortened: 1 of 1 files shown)


src/main/java/io/cdap/e2e/pages/actions/CdfStudioActions.java
```diff
@@ -309,7 +309,10 @@ public static String bannerErrorMessage() {
    * Click on the Plugin group title: Sink
    */
   public static void clickSink() {
-    ElementHelper.clickOnElement(CdfStudioLocators.sink);
+    // Check if Sink is not already expanded
+    if (!ElementHelper.isElementDisplayed(CdfStudioLocators.bigQueryObject)) {
```
itsankit-google, line 313: I think it would be better to add another method like "expandSinkPluginGroup" and then add this check. This method is only meant for clicking on the sink group and not ensuring whether it is expanded.

mrahanjam: Sounds good, in that case all the `When sink is ...` steps should be updated to use that method.

PR #156, opened 2022-11-09
set default download path to target for chrome driver

mrahanjam:
This PR sets the default download path for chrome driver to `target/downloads` directory. It helps with the tests where a file (e.g., exported pipeline json) needs to be downloaded and re-read from the disk.


itsankit-google (approved):
lgtm

PR #172, opened 2023-03-31
Updating sandbox url to fix Github Action failures.

Vipinofficial11:
Github Actions are not able to download the CDAP sandbox using the current link.

### DIFF
--- a/src/main/scripts/run_e2e_test.py
+++ b/src/main/scripts/run_e2e_test.py
@@ -40,7 +40,7 @@ args=parser.parse_args()
 
 # Start CDAP sandbox
 print("Downloading CDAP sandbox")
-sandbox_url = "https://github.com/cdapio/cdap-build/releases/download/latest/cdap-sandbox-6.9.0-SNAPSHOT.zip"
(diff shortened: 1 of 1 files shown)


src/main/scripts/run_e2e_test.py
```diff
@@ -40,7 +40,7 @@ def run_shell_command(cmd):
 
 # Start CDAP sandbox
 print("Downloading CDAP sandbox")
-sandbox_url = "https://github.com/cdapio/cdap-build/releases/download/latest/cdap-sandbox-6.9.0-SNAPSHOT.zip"
+sandbox_url = "https://github.com/cdapio/cdap-build/releases/download/v6.9.0-SNAPSHOT/cdap-sandbox-6.9.0-SNAPSHOT.zip"
```
de-lan: Should it instead be `/latest/cdap-sandbox-6.10.0-SNAPSHOT.zip`?

Vipinofficial11: yes it should have been ``` /latest/cdap-sandbox-6.10.0-SNAPSHOT.zip ```. There is an another PR for this same change that has the snapshot version 6.10.0 so closing this one. (PR : https://github.com/cdapio/cdap-e2e-tests/pull/173)

PR #179, opened 2023-05-29
Added actions and locators for BQ sink Table Key and Dedupe By

Aryan-Verma:
Added actions and locators for BigQuery Sink Table Key and Dedupe By properties.


MrRahulSharma (approved):
LGTM

PR #180, opened 2023-05-30
add cdp 113 dependency

GnsP:


pom.xml
```diff
@@ -369,6 +369,13 @@ the License.-->
       <version>2.1.8</version>
     </dependency>
 
+    <!-- https://mvnrepository.com/artifact/org.seleniumhq.selenium/selenium-devtools-v113 -->
+    <dependency>
+      <groupId>org.seleniumhq.selenium</groupId>
+      <artifactId>selenium-devtools-v113</artifactId>
```
itsankit-google, line 375: why is this dependency required?

GnsP: It was the issue with my local build. Looks like the issue on the github runner is something else.

PR #195, opened 2023-08-11
Adding parameter for multiple profiles

itsmekumari:
Adding parameter for multiple profiles. For database-delta-plugins repo the build was failing due to docker-maven which would not be required for e2e tests. Along with the e2e-tests which is default profile, it also required the skip-its profile to build the module. This change added as parameter to use multiple profiles.

### DIFF
--- a/src/main/scripts/run_e2e_test.py
+++ b/src/main/scripts/run_e2e_test.py
@@ -1,4 +1,4 @@
-# Copyright © 2021 Cask Data, Inc.
+# Copyright © 2023 Cask Data, Inc.
 #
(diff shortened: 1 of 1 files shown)


itsankit-google (changes requested):
Build is failing with error:
```
Traceback (most recent call last):
  File "e2e/src/main/scripts/run_e2e_test.py", line 148, in <module>
    if args.mvnProfile:
AttributeError: 'Namespace' object has no attribute 'mvnProfile'
```

Can we please make sure build status is verified to be green before sending the PR for review.
cc: @poojantcs @Vipinofficial11

itsankit-google (approved):

PR #197, opened 2023-08-21
CDF Login using OTA 

Vipinofficial11:
Changes for CDF login flow using OTA credentials.

### DIFF
--- a/src/main/java/io/cdap/e2e/pages/actions/CdfPipelineRunAction.java
+++ b/src/main/java/io/cdap/e2e/pages/actions/CdfPipelineRunAction.java
@@ -21,6 +21,7 @@ import io.cdap.e2e.utils.AssertionHelper;
 import io.cdap.e2e.utils.ConstantsUtil;
 import io.cdap.e2e.utils.ElementHelper;
 import io.cdap.e2e.utils.PageHelper;
+import io.cdap.e2e.utils.RetryUtils;
 import io.cdap.e2e.utils.SeleniumDriver;
 import io.cdap.e2e.utils.SeleniumHelper;
 import io.cdap.e2e.utils.WaitHelper;
@@ -86,6 +87,15 @@ public class CdfPipelineRunAction {
   }
 
   /**
+   * Check if the pipeline is in Starting Status
+   *
+   * @return Boolean
+   */
+  public static Boolean isStarting() {
+    return validatePipelineExpectedStatus(CdfPipelineRunLocators.startingStatus);
+  }
+
+  /**
    * Check if the pipeline is in Provisioning Status
    *
    * @return Boolean
@@ -107,14 +117,18 @@ public class CdfPipelineRunAction {
 
   /**
    * Wait till the Pipeline's status changes (from Running) to either Succeeded, Failed or Stopped within the
-   * Timeout: {@link ConstantsUtil#PIPELINE_RUN_TIMEOUT_SECONDS}
+   * Timeout: {@link ConstantsUtil#IMPLICIT_TIMEOUT_SECONDS}
    */
-  public static void waitTillPipelineRunCompletes() {
-    SeleniumDriver.getWaitDriver(ConstantsUtil.PIPELINE_RUN_TIMEOUT_SECONDS).until(ExpectedConditions.or(
-      ExpectedConditions.visibilityOf(CdfPipelineRunLocators.succeededStatus),
-      ExpectedConditions.visibilityOf(CdfPipelineRunLocators.failedStatus),
-      ExpectedConditions.visibilityOf(CdfPipelineRunLocators.stoppedStatus)
-    ));
+  public static void waitTillPipelineRunCompletes() throws InterruptedException {
+    // Adding a page refresh in case tests are running on CDF to update the pipeline status.
+      RetryUtils.retry(ConstantsUtil.PIPELINE_REFRESH_TIMEOUT_SECONDS, ConstantsUtil.PIPELINE_RUN_TIMEOUT_SECONDS,
+        10, () -> {
+        PageHelper.refreshCurrentPage();
+        return !(isRunning());
+        }
+      );
+
+    waitTillPipelineRunCompletes(ConstantsUtil.IMPLICIT_TIMEOUT_SECONDS);
   }
 
   /**
--- a/src/main/java/io/cdap/e2e/pages/actions/CdfSignInActions.java
+++ b/src/main/java/io/cdap/e2e/pages/actions/CdfSignInActions.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright © 2023 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.cdap.e2e.pages.actions;
+
+import io.cdap.e2e.pages.locators.CdfSignInLocator;
+import io.cdap.e2e.utils.ConstantsUtil;
+import io.cdap.e2e.utils.ElementHelper;
+import io.cdap.e2e.utils.SeleniumHelper;
+import io.cdap.e2e.utils.WaitHelper;
+
+import java.io.IOException;
+
+
+
+/**
+ * Represents CdfSignInActions
+ */
+public class CdfSignInActions {
+    private static CdfSignInLocator cdfSignInLocator;
+    static {
+        cdfSignInLocator = SeleniumHelper.getPropertiesLocators(CdfSignInLocator.class);
+    }
+
+    public static void login() throws IOException, InterruptedException {
+        ElementHelper.sendKeys(cdfSignInLocator.cdfUsername, SeleniumHelper.readParameters(ConstantsUtil.CDFUSERNAME));
+        ElementHelper.clickOnElement(cdfSignInLocator.nextButton);
+        ElementHelper.sendKeys(cdfSignInLocator.cdfPassword, SeleniumHelper.readParameters(ConstantsUtil.CDFPASSWORD));
+        ElementHelper.clickOnElement(cdfSignInLocator.nextButton);
+
+        ElementHelper.clickIfDisplayed(cdfSignInLocator.selectTestAccount(ConstantsUtil.CDF_TEST_ACCOUNT_NAME),
+          ConstantsUtil.SMALL_TIMEOUT_SECONDS, cdfSignInLocator.clickOnContinueButton());
+
+        ElementHelper.clickIfDisplayed(cdfSignInLocator.clickOnContinueButton(), ConstantsUtil.SMALL_TIMEOUT_SECONDS,
+          cdfSignInLocator.locatePluginNameInList(ConstantsUtil.FIRST_PLUGIN_IN_LIST, "Source"));
+
+        ElementHelper.clickIfDisplayed(cdfSignInLocator.clickOnAllowButton(), ConstantsUtil.SMALL_TIMEOUT_SECONDS,
+          cdfSignInLocator.locatePluginNameInList(ConstantsUtil.FIRST_PLUGIN_IN_LIST, "Source"));
+
+    }
+
+    public static boolean isUserLoggedInCDF() {
+        return !WaitHelper.waitForElementToBeOptionallyDisplayed(
+                CdfSignInLocator.locatorOfEmailTextBox(), ConstantsUtil.SMALL_TIMEOUT_SECONDS);
+    }
+}
--- a/src/main/java/io/cdap/e2e/pages/locators/CdfSignInLocator.java
+++ b/src/main/java/io/cdap/e2e/pages/locators/CdfSignInLocator.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright © 2023 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.cdap.e2e.pages.locators;
+
+import org.openqa.selenium.By;
+import org.openqa.selenium.WebElement;
+import org.openqa.selenium.support.FindBy;
+import org.openqa.selenium.support.How;
+
+/**
+ * Represents CdfSignInLocators
+ */
+public class CdfSignInLocator {
+
+    @FindBy(how = How.XPATH, using = "//input[@name='identifier']")
+    public WebElement cdfUsername;
+
+    @FindBy(how = How.XPATH, using = "//input[@name='Passwd']")
+    public WebElement cdfPassword;
+
+    @FindBy(how = How.XPATH, using = "//*[contains(text(),'Next')]//parent::button")
+    public WebElement nextButton;
+
+    public static By locatorOfEmailTextBox() {
+        return By.xpath("//input[@name='identifier']");
+    }
+
+    public By selectTestAccount(String testAccountName) {
+        return By.xpath("//*[contains(text(),'" + testAccountName + "')]");
+    }
+
+    public By clickOnAllowButton() {
+        return By.xpath("//button[contains(text(),'Allow')]");
+    }
+
+    public By clickOnContinueButton() {
+        return By.xpath("//span[contains(text(),'Continue')]");
+    }
+
+    public static By locatePluginNameInList(String pluginName, String  pluginGroupName) {
+        return By.xpath("//div[@data-cy='plugin-" + pluginGroupName + "-group']" +
+                "//div[contains(@class, 'PluginNameContainer')][normalize-space(text()) = '" + pluginName + "' " +
+                "or translate(normalize-space(text()),' ','') = '" + pluginName + "']");
+    }
+}
--- a/src/main/java/io/cdap/e2e/utils/ConstantsUtil.java
+++ b/src/main/java/io/cdap/e2e/utils/ConstantsUtil.java
@@ -34,6 +34,10 @@ public class ConstantsUtil {
   public static final String GPAZWRD = "gPassword";
   public static final String HDFUSERNAME = "hdfUsername";
   public static final String HDFPAZWRD = "hdfPassword";
+  public static final String CDFUSERNAME = "cdfUsername";
+  public static final String CDFPASSWORD = "cdfPassword";
+  public static final String TESTONCDF = "testOnCdf";
+  public static final String CDF_TEST_ACCOUNT_NAME = "CloudDataFution Automation";
   public static final String CDFURL = "cdfurl";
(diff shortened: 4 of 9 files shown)


src/main/resources/connectionParameters.properties
```diff
@@ -9,7 +9,12 @@ testOnHdf=false
 hdfUsername=cdap
 hdfPassword=cdap123
 
+testOnCdf=false
+cdfUsername=dummy
+cdfPassword=dummy
+
 screenshotForAllSteps=false
+testAccountName=CloudDataFution Automation
```
rmstar: nit: Fution->Fusion

src/main/java/io/cdap/e2e/pages/actions/CdfPipelineRunAction.java
```diff
@@ -109,7 +119,16 @@ public static void waitForPipelineToTransitionToStatus(String pipelineStatus, lo
    * Wait till the Pipeline's status changes (from Running) to either Succeeded, Failed or Stopped within the
    * Timeout: {@link ConstantsUtil#PIPELINE_RUN_TIMEOUT_SECONDS}
    */
-  public static void waitTillPipelineRunCompletes() {
+  public static void waitTillPipelineRunCompletes() throws InterruptedException, IOException {
+    // Adding a page refresh in case tests are running on CDF to update the pipeline status.
+    if (Boolean.parseBoolean(SeleniumHelper.readParameters(ConstantsUtil.TESTONCDF)) ||
```
rmstar: Is there a reason why we only want to refresh the page when running on CDF and HDF?

src/main/resources/connectionParameters.properties
Vipinofficial11: I am not sure why, but it actually shows ```CloudDataFution Automation``` on the UI

src/main/java/io/cdap/e2e/pages/actions/CdfPipelineRunAction.java
Vipinofficial11: Yes, I encountered a bug a lot of times on CDF and HDF where the pipeline status gets stuck and doing a page refresh or clicking somewhere on the UI updates the pipeline status.

haven't seen this on CDAP test executions so far.

rmstar (approved):

src/main/java/io/cdap/e2e/pages/actions/CdfSignInActions.java
```diff
@@ -0,0 +1,66 @@
...
+package io.cdap.e2e.pages.actions;
+
+import io.cdap.e2e.pages.locators.CdfSignInLocator;
+import io.cdap.e2e.utils.ConstantsUtil;
+import io.cdap.e2e.utils.ElementHelper;
+import io.cdap.e2e.utils.SeleniumHelper;
+import io.cdap.e2e.utils.WaitHelper;
+import java.io.IOException;
+import static io.cdap.e2e.utils.ConstantsUtil.CDFPASSWORD;
+import static io.cdap.e2e.utils.ConstantsUtil.CDFUSERNAME;
```
itsankit-google: unless there is a specific usecase, we should not go for static imports, it makes the code unreadable.

Please use `ConstantsUtil.CDFPASSWORD` and so on wherever required.

sau42shri: @Vipinofficial11 can you address comment so that we can get to a closure sooner.

Vipinofficial11: updated [36bc9cff025b14bc9e7b456af4d9fa4b87c42082]

src/main/java/io/cdap/e2e/pages/locators/CdfSignInLocator.java
```diff
@@ -0,0 +1,58 @@
...
+
+    public By clickOnAllowButton() {
+        return By.xpath("//button[contains(text(),'Allow')]");
+    }
+
+    public By clickOnContinueButton() {
+        return By.xpath("//span[contains(text(),'Continue')]");
+    }
+
+    public static By locatePluginNameInList(String pluginName, String  pluginGroupName) {
```
itsankit-google, line 53: This locator is already present in `CdfStudioLocators`, can we reuse it??

src/main/java/io/cdap/e2e/pages/actions/CdfPipelineRunAction.java
```diff
@@ -109,7 +119,16 @@ public static void waitForPipelineToTransitionToStatus(String pipelineStatus, lo
...
+      Boolean.parseBoolean(SeleniumHelper.readParameters(ConstantsUtil.TESTONHDF))) {
+      do {
+        PageHelper.refreshCurrentPage();
+        SeleniumDriver.getWaitDriver(ConstantsUtil.PAGE_LOAD_TIMEOUT_SECONDS);
```
itsankit-google: A page refresh in every `50seconds` might be too frequent, it won't change the pipeline status drastically, I would suggest adding a sleep of atleast `2 mins`.

src/main/java/io/cdap/e2e/pages/locators/CdfSignInLocator.java
Vipinofficial11: The Locator present in ```CdfStudioLocators``` has return type as WebElement while this one has a return type of By because the use case for both is a bit different.

1. The ```locatePluginNameInList```  in ```CdfStudioLocators``` is being used to check if we are able to find a plugin in the plugin list. (Here we are using this with ```WaitHelper.waitForElementToBeDisplayed```)

2. The ```locatePluginNameInList``` in ```CdfSingInLocators``` is being used in sort of a retry mechanism where we are checking if after clicking on the Continue button we are able to see the studio page or a plugin in the plugin list.  (Here we are using this with ```WaitHelper.waitForElementToBeOptionallyDisplayed```)

The locator surely should be moved to ```CdfStudioLocators``` but I don't see a way of reusing the one whose return type is WebElement. 

What do you think, is there any way using which we can reuse the locator and achieve both use cases.

src/main/java/io/cdap/e2e/pages/actions/CdfSignInActions.java
```diff
@@ -0,0 +1,65 @@
...
+
+        String testAccountName = SeleniumHelper.readParameters("testAccountName");
+        ElementHelper.clickIfDisplayed(cdfSignInLocator.selectTestAccount(testAccountName),
+          ConstantsUtil.MEDIUM_TIMEOUT_SECONDS, cdfSignInLocator.clickOnAllowButton());
+
+        ElementHelper.clickIfDisplayed(cdfSignInLocator.clickOnContinueButton(), ConstantsUtil.MEDIUM_TIMEOUT_SECONDS,
+          cdfSignInLocator.locatePluginNameInList(ConstantsUtil.FIRST_PLUGIN_IN_LIST, "Source"));
+
+        ElementHelper.clickIfDisplayed(cdfSignInLocator.clickOnAllowButton(), ConstantsUtil.MEDIUM_TIMEOUT_SECONDS,
+          cdfSignInLocator.locatePluginNameInList(ConstantsUtil.FIRST_PLUGIN_IN_LIST, "Source"));
```
itsankit-google: Why do we want to combine two operations here, why can't we do something like this?
1. Click on `element1` if displayed
2. Click on `element2` if displayed
3. Wait until `plugin element` is displayed.

src/main/java/io/cdap/e2e/pages/locators/CdfSignInLocator.java
itsankit-google: Commented: https://github.com/cdapio/cdap-e2e-tests/pull/197#discussion_r1307602850

src/main/java/io/cdap/e2e/pages/actions/CdfSignInActions.java
Vipinofficial11: The main reason behind creating ```clickIfDisplayed``` was to implement a retry if the element is not displayed on the first attempt. 

The reason to combine 2 operations was related to this retry only, the idea was that if once we clicked on the desired element we will validate if the next expected element (3rd parameter) is displayed if we don't see the next element we try to click on the desired element again.

itsankit-google: Will this retry be implemented in future? I don't see it being retried in this PR or am I missing something?

Vipinofficial11: The Retry logic is actually implemented in ```clickIfDisplayed``` 

```

    clickIfDisplayed(locatorToClick, timeOutInSeconds);

    // If the next webElement is not displayed we will click on the desired element again.
    if (!isElementDisplayed(locatorToWaitFor, timeOutInSeconds)) {
    logger.info("Retry : Click on " + locatorToClick);
    clickIfDisplayed(locatorToClick);
      
```


(https://github.com/cdapio/cdap-e2e-tests/pull/197/files#diff-1b63d2aab0be21d4b99280309e329d962acb6a5c2bb82e6075c8da4ce3b6569eR134)

src/main/java/io/cdap/e2e/pages/actions/CdfSignInActions.java
```diff
@@ -0,0 +1,65 @@
...
+        cdfSignInLocator = SeleniumHelper.getPropertiesLocators(CdfSignInLocator.class);
+    }
+
+    public static void login() throws IOException {
+        ElementHelper.sendKeys(cdfSignInLocator.cdfUsername, SeleniumHelper.readParameters(ConstantsUtil.CDFUSERNAME));
+        ElementHelper.clickOnElement(cdfSignInLocator.nextButton);
+        ElementHelper.sendKeys(cdfSignInLocator.cdfPassword, SeleniumHelper.readParameters(ConstantsUtil.CDFPASSWORD));
+        ElementHelper.clickOnElement(cdfSignInLocator.nextButton);
+
+        // Adding optional operations here.
```
samdgupi: what is this comment referring to ?

src/main/java/io/cdap/e2e/pages/actions/CdfSignInActions.java
```diff
@@ -0,0 +1,65 @@
...
+
+        /*
+        Based on the observations in case of CDF test execution from local we encounter Allow Button
+        and in case of CDF test execution from Cloud Build we encounter Continue Button that's why we
+        are using 2 confirmation buttons here.
+        */
+
+        String testAccountName = SeleniumHelper.readParameters("testAccountName");
```
samdgupi: Please define constant for "testAccountName"

src/main/java/io/cdap/e2e/utils/ConstantsUtil.java
```diff
@@ -93,4 +96,9 @@ public class ConstantsUtil {
    * PIPELINE_RUN_TIMEOUT_SECONDS: To be used as a timeout for Pipeline Runs
    */
   public static final int PIPELINE_RUN_TIMEOUT_SECONDS = 900;
+
+  /**
+   * MEDIUM_TIMEOUT_SECONDS: To be used as a wait for retry
```
samdgupi: Can you please make the comment and name specific. 
This is used to wait for elemnt to be displayed right ?

src/main/java/io/cdap/e2e/utils/ConstantsUtil.java
```diff
@@ -93,4 +96,9 @@ public class ConstantsUtil {
...
+   */
+  public static final int MEDIUM_TIMEOUT_SECONDS = 10;
```
samdgupi: Looking at the usage would not SMALL_TIMEOUT_SECONDS work ?

src/main/java/io/cdap/e2e/pages/actions/CdfSignInActions.java
samdgupi: why is the retry needed? The click should result in the next element within timeout. If we retry, we might initiate the same operation twice.

src/main/java/io/cdap/e2e/pages/actions/CdfPipelineRunAction.java
```diff
@@ -109,7 +119,16 @@ public static void waitForPipelineToTransitionToStatus(String pipelineStatus, lo
...
+      } while (isStarting() || isRunning() || isProvisioning());
+    }
+
     SeleniumDriver.getWaitDriver(ConstantsUtil.PIPELINE_RUN_TIMEOUT_SECONDS).until(ExpectedConditions.or(
       ExpectedConditions.visibilityOf(CdfPipelineRunLocators.succeededStatus),
```
samdgupi: This wait seems unneeded based on the above retry which will only stop when the pipeline is in a terminal state

src/main/java/io/cdap/e2e/pages/actions/CdfSignInActions.java
Vipinofficial11: Created a constant ```CDF_TEST_ACCOUNT_NAME``` for this.  

(https://github.com/cdapio/cdap-e2e-tests/pull/197/commits/44be5d5c5c9bf2c3ee877d3eb7dd3aa43c414929#diff-d1ebf05f182d1decee9527cecadb7ea940ba7cc8dddc7519c80dd3e4b64db94fR40)

src/main/java/io/cdap/e2e/utils/ConstantsUtil.java
Vipinofficial11: The small timeout works fine here, so removed this Timeout.

src/main/java/io/cdap/e2e/utils/ConstantsUtil.java
Vipinofficial11: It works on local execution, used  ```SMALL_TIMEOUT_SECONDS``` instead of medium everywhere for now. 

Note : There is a possibility that due to some external delay in the Login Flow a small timeout may cause some flakiness.

(https://github.com/cdapio/cdap-e2e-tests/pull/197/commits/44be5d5c5c9bf2c3ee877d3eb7dd3aa43c414929#diff-653a26322c5148b982b3df96f0640a19488ea600b8501ce2695a5218dda8be2bR45)

src/main/java/io/cdap/e2e/pages/actions/CdfSignInActions.java
Vipinofficial11: Based on the observations in case of CDF test execution from local we encountered an Allow Button and in case of CDF test execution from Cloud Build we encountered Continue Button so to handle this scenario the button clicks were optional. 

That's not the case now I see same Continue button on both executions now, so removed the Allow button click and comment as well.

src/main/java/io/cdap/e2e/pages/actions/CdfSignInActions.java
Vipinofficial11: I agree that this can result in same operation twice, the main purpose of adding the refresh here was to avoid any flakiness in case of CDF login flow because I observed few failed executions while executing the flow. I am not completely sure why but sometimes the click was getting intercepted. (We were waiting for elementToBeClickable still there were some failed executions due to this)

We can remove this refresh and use a simple click with a timeout, I will just have to make sure that there is no flakiness while using a simple click on Cloud Build.

what do you guys suggest ? (cc : @itsankit-google)

src/main/java/io/cdap/e2e/pages/actions/CdfPipelineRunAction.java
Vipinofficial11: Updated the sleep time to 2mins, re-using ```PIPELINE_DEPLOY_TIMEOUT_SECONDS``` as it's value is 120, but the constant name can create some confusion here.

Should we create a new constant with same value named as ```PIPELINE_REFRESH_TIMEOUT_SECONDS``` here ?  

(https://github.com/cdapio/cdap-e2e-tests/pull/197/commits/44be5d5c5c9bf2c3ee877d3eb7dd3aa43c414929#diff-cfa45894e4592a5d6ed32f485ed2594ae01b9a3b154e60820a599164e0b52e80R136)

Made some changes to this function PTAL

src/main/java/io/cdap/e2e/pages/actions/CdfPipelineRunAction.java
Vipinofficial11: Agree, The purpose of this wait here was if the pipeline is in Running status for more than ```15 minutes```  we should mark that pipeline as failed to avoid a very long Build execution time. 

Now I realised that after the above refresh logic, the do-while loop keep on refreshing the page until the pipeline is in running state which is a problem what if pipeline takes 50-60 mins. 

So added a logic in the do-while to break the loop if the pipeline is in running state for more that ```900second or 15 mins``` also changed the ```PIPELINE_RUN_TIMEOUT_SECONDS = 900second``` to ```IMPLICIT_TIMEOUT_SECONDS = 30second``` to check if  ```Failed || Success || Stop``` status is seen.

cc: @itsankit-google 

(https://github.com/cdapio/cdap-e2e-tests/pull/197/commits/44be5d5c5c9bf2c3ee877d3eb7dd3aa43c414929#diff-cfa45894e4592a5d6ed32f485ed2594ae01b9a3b154e60820a599164e0b52e80R122)

src/main/java/io/cdap/e2e/pages/actions/CdfPipelineRunAction.java
```diff
@@ -107,10 +117,27 @@ public static void waitForPipelineToTransitionToStatus(String pipelineStatus, lo
...
-    SeleniumDriver.getWaitDriver(ConstantsUtil.PIPELINE_RUN_TIMEOUT_SECONDS).until(ExpectedConditions.or(
+  public static void waitTillPipelineRunCompletes() throws InterruptedException, IOException {
+    int pipelineExecutionTimeFlag = 0;
+    // Adding a page refresh in case tests are running on CDF to update the pipeline status.
+    if (Boolean.parseBoolean(SeleniumHelper.readParameters(ConstantsUtil.TESTONCDF)) ||
+      Boolean.parseBoolean(SeleniumHelper.readParameters(ConstantsUtil.TESTONHDF))) {
+      // Adding pipelineExecutionTimeFlag to break the loop if pipeline status is Running state for more than
+      // 900 seconds.
+      do {
+        pipelineExecutionTimeFlag += 120;
```
samdgupi: Can we please define a retry function which takes a retry delay, max retry delay and max retrycount. This can be used both here and in clickIfDisplayed

samdgupi: Also please make the retry delays as constants

Vipinofficial11: Added the retry function and re-used it for both pipeline refresh and button click. (5c54894de6f63366c43ab25cc9655e52f9c1d468)

Vipinofficial11:
Current Build Failures are not related to these changes. Team is working on fixing these failures.

Screenshot 2023-10-03 at 10 40 12 AM

src/main/java/io/cdap/e2e/pages/actions/CdfPipelineRunAction.java
```diff
@@ -107,10 +122,20 @@ public static void waitForPipelineToTransitionToStatus(String pipelineStatus, lo
...
+      RetryUtils.retry(ConstantsUtil.PIPELINE_REFRESH_TIMEOUT_SECONDS, ConstantsUtil.PIPELINE_RUN_TIMEOUT_SECONDS,
+        10, () -> {
+        logger.info("Refreshing the page.");
+        PageHelper.refreshCurrentPage();
+        SeleniumDriver.getWaitDriver(ConstantsUtil.SMALL_TIMEOUT_SECONDS);
+        return !(isStarting() || isRunning() || isProvisioning());
+        }
+      );
+
+    SeleniumDriver.getWaitDriver(ConstantsUtil.IMPLICIT_TIMEOUT_SECONDS).until(ExpectedConditions.or(
```
samdgupi: This looks similar to `waitTillPipelineRunCompletes`. Please reuse.

Vipinofficial11: Re-used (66e7a8071c28508450e1813edfeda2ccb7be0f60)

samdgupi (approved):

Vipinofficial11:
Merging the PR as it is approved by 2 reviewers (Tested these changes on CDAP as well as on CDF setup)

PR #212, opened 2023-09-21
Adding scripts to deploy required plugins on sandbox from CDAP hub.

Vipinofficial11:
Added a script to deploy required plugins from Hub which are not present by default in CDAP studio. Using the Hub API for the Plugin deployment. [(Hub API)](https://cdap.atlassian.net/wiki/spaces/DOCS/pages/554401840/Hub+API)

Flow :-

1. Add the plugin name and version in ``` required_plugins.yaml``` which you want to deploy on the sandbox.
2. Once the sandbox is created we read the plugin name and plugin version added in ``` required_plugins.yaml``` and the ``` upload_required_plugins.py ``` script is called which will upload each plugin one by one on default namespace.

### DIFF
--- a/src/main/scripts/upload_required_plugins.py
+++ b/src/main/scripts/upload_required_plugins.py
@@ -30,10 +30,13 @@ actions = response_data["actions"][0]
 plugin_details = actions["arguments"]
 
 # Get the required details from specs.
-plugin_name = plugin_details[0]["value"]
-plugin_version = plugin_details[1]["value"]
-plugin_config = plugin_details[3]["value"]
-plugin_jar = plugin_details[4]["value"]
+plugin_details_dictionary = {
+    item['name']: item['value'] for item in plugin_details
+}
+plugin_name = plugin_details_dictionary['name']
+plugin_version = plugin_details_dictionary['version']
+plugin_config = plugin_details_dictionary['config']
+plugin_jar = plugin_details_dictionary['jar']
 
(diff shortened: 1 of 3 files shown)


src/main/scripts/upload_required_plugins.py
```diff
@@ -0,0 +1,68 @@
...
+assert res.ok or print(f"Specs not found for the Plugin. Status code: {res.status_code}")
+response_data = res.json()
+actions = response_data["actions"][0]
+plugin_details = actions["arguments"]
+
+# Get the required details from specs.
+plugin_name = plugin_details[0]["value"]
+plugin_version = plugin_details[1]["value"]
+plugin_config = plugin_details[3]["value"]
+plugin_jar = plugin_details[4]["value"]
```
GnsP: Rely on the `name` fields in the `arguments` array elements, instead of list indices.

GnsP (changes requested):

src/main/scripts/upload_required_plugins.py
Vipinofficial11: Updated, used the name fields to extract the specs. [1d4cda1ac469839b331b485353f5032cc5f1a469]

GnsP (approved):

Vipinofficial11:
Screenshot 2023-10-05 at 7 37 38 PM

2 tests failing due to intermittent issue, merging the PR as it is approved.

PR #218, opened 2023-10-12
e2e tests build failures fix

itsmekumari:
e2e tests build failure fix changes for older flaky scenarios in Google Cloud and Hydrator repos. For Reference https://cdap.atlassian.net/browse/PLUGIN-1693

### DIFF
--- a/src/main/java/io/cdap/e2e/pages/actions/CdfPluginPropertiesActions.java
+++ b/src/main/java/io/cdap/e2e/pages/actions/CdfPluginPropertiesActions.java
@@ -141,10 +141,15 @@ public class CdfPluginPropertiesActions {
    */
   public static void clickGetSchemaButton() {
     ElementHelper.clickOnElement(CdfPluginPropertiesLocators.getSchemaButton);
+    ElementHelper.elementToBeOptionallyDisabled(CdfPluginPropertiesLocators.getSchemaButton);
+    /*
+    TODO : Remove the commented lines of code once the bug is resolved by Chrome Team
+    Bug : https://cdap.atlassian.net/browse/PLUGIN-1693
     WaitHelper.waitForElementToBeOptionallyDisplayed(
       CdfPluginPropertiesLocators.locatorOfLoadingSpinnerOnGetSchemaButton(), ConstantsUtil.SMALL_TIMEOUT_SECONDS);
     WaitHelper.waitForElementToBeHidden(
       CdfPluginPropertiesLocators.locatorOfLoadingSpinnerOnGetSchemaButton(), ConstantsUtil.DEFAULT_TIMEOUT_SECONDS);
+     */
     WaitHelper.waitForElementToBeDisplayed(CdfPluginPropertiesLocators.getSchemaButton);
   }
 
@@ -153,10 +158,15 @@ public class CdfPluginPropertiesActions {
    */
   public static void clickValidateButton() {
     ElementHelper.clickOnElement(CdfPluginPropertiesLocators.validateButton);
+    ElementHelper.elementToBeOptionallyDisabled(CdfPluginPropertiesLocators.validateButton);
+    /*
+    TODO : Remove the commented lines of code once the bug is resolved by Chrome Team
+    Bug : https://cdap.atlassian.net/browse/PLUGIN-1693
     WaitHelper.waitForElementToBeOptionallyDisplayed(
       CdfPluginPropertiesLocators.locatorOfLoadingSpinnerOnValidateButton(), ConstantsUtil.SMALL_TIMEOUT_SECONDS);
     WaitHelper.waitForElementToBeHidden(
       CdfPluginPropertiesLocators.locatorOfLoadingSpinnerOnValidateButton(), ConstantsUtil.DEFAULT_TIMEOUT_SECONDS);
+   */
     WaitHelper.waitForElementToBeDisplayed(CdfPluginPropertiesLocators.validateButton);
   }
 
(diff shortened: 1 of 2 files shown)


src/main/java/io/cdap/e2e/pages/actions/CdfPluginPropertiesActions.java
```diff
@@ -141,23 +141,33 @@ public static void changeDataTypesInOutputSchema(String jsonFieldList) {
...
   }
 
   /**
    * Click on the Validate button inside Plugin's properties page
    */
   public static void clickValidateButton() {
     ElementHelper.clickOnElement(CdfPluginPropertiesLocators.validateButton);
+    WaitHelper.waitForElementToBeDisabled(CdfPluginPropertiesLocators.validateButton);
+    /*
+    TODO : Remove the commented lines of code once the bug is resolved by Chrome Team.
```
samdgupi: Is there a chome bug link ? How are we sure this is a chrome issue ?

rahuldash171: https://bugs.chromium.org/p/chromedriver/issues/detail?id=4440 
https://cdap.atlassian.net/browse/PLUGIN-1693

Can you check both the issues , one is raised for chrome browser and recently we have raise the second one for CDAP where the error is same "No node with given id found" .

src/main/java/io/cdap/e2e/utils/WaitHelper.java
```diff
@@ -267,4 +267,20 @@ public static boolean waitForTextToBePresentInElementValue(WebElement element, S
                   " with the Default timeout: " + ConstantsUtil.DEFAULT_TIMEOUT_SECONDS + " seconds");
     return SeleniumDriver.getWaitDriver().until(ExpectedConditions.textToBePresentInElementValue(element, text));
   }
+
+  /**
+   * Wait for the element to be disabled
+   *
+   * @param element WebElement to check whether click is disabled for temporary time
+   */
+  public static boolean waitForElementToBeDisabled(WebElement element) {
```
samdgupi: I think this should be similar to `waitForElementToBeEnabled`. Only the `return element.isEnabled()` should be `return !element.isEnabled()`

src/main/java/io/cdap/e2e/utils/WaitHelper.java
```diff
@@ -267,4 +267,20 @@ public static boolean waitForTextToBePresentInElementValue(WebElement element, S
...
+    logger.info("Waiting for the element: " + element + " to be disabled ");
+    try {
+      SeleniumDriver.getWaitDriver().until(ExpectedConditions.elementToBeClickable(element));
```
samdgupi: I don't think this is semantically correct based on what the function is expected to do

src/main/java/io/cdap/e2e/pages/actions/CdfPluginPropertiesActions.java
```diff
@@ -141,23 +141,33 @@ public static void changeDataTypesInOutputSchema(String jsonFieldList) {
...
     WaitHelper.waitForElementToBeDisplayed(CdfPluginPropertiesLocators.getSchemaButton);
+    */
   }
 
   /**
    * Click on the Validate button inside Plugin's properties page
    */
   public static void clickValidateButton() {
     ElementHelper.clickOnElement(CdfPluginPropertiesLocators.validateButton);
+    WaitHelper.waitForElementToBeDisabled(CdfPluginPropertiesLocators.validateButton);
```
samdgupi: We should follow this up with `WaitHelper.waitForElementToBeEnabled`
essentially checking click->disabled(gif)->enabled

samdgupi (changes requested):

src/main/java/io/cdap/e2e/utils/WaitHelper.java
rahuldash171: Yes Samik , we have change the method name to **elementToBeOptionallyDisabled** and returning the same whether the element is disabled optianally.  We have removed the wait as the check is to verify if the element is disabled after clicked .

src/main/java/io/cdap/e2e/utils/WaitHelper.java
rahuldash171: We have changed the method contents as discussed , please see the below file .
 public static boolean elementToBeOptionallyDisabled(WebElement element) {
    logger.info("The element: " + element + " is disabled ");
    return !element.isEnabled();

src/main/java/io/cdap/e2e/pages/actions/CdfPluginPropertiesActions.java
rahuldash171: yes now we have changed it to ElementHelper.elementToBeOptionallyDisabled . Here we are just checking if the element is disabled before displayed and clickable again .

src/main/java/io/cdap/e2e/pages/actions/CdfPluginPropertiesActions.java
```diff
@@ -153,10 +158,15 @@ public static void clickGetSchemaButton() {
...
+    /*
+    TODO : Remove the commented lines of code once the bug is resolved by Chrome Team.
+    Bug : https://cdap.atlassian.net/browse/PLUGIN-1693
     WaitHelper.waitForElementToBeOptionallyDisplayed(
       CdfPluginPropertiesLocators.locatorOfLoadingSpinnerOnValidateButton(), ConstantsUtil.SMALL_TIMEOUT_SECONDS);
     WaitHelper.waitForElementToBeHidden(
       CdfPluginPropertiesLocators.locatorOfLoadingSpinnerOnValidateButton(), ConstantsUtil.DEFAULT_TIMEOUT_SECONDS);
+   */
     WaitHelper.waitForElementToBeDisplayed(CdfPluginPropertiesLocators.validateButton);
```
samdgupi, line 170: I think we should check if it is enabled again. Displayed might always be true

samdgupi (approved):

PR #229, opened 2023-11-15
Functionality added for GCS multipart upload policy.

neerajsinghal05:
Functionality added for GCS multipart upload policy.

### DIFF
--- a/src/main/java/io/cdap/e2e/utils/StorageClient.java
+++ b/src/main/java/io/cdap/e2e/utils/StorageClient.java
@@ -22,10 +22,15 @@ import com.google.cloud.storage.BlobId;
 import com.google.cloud.storage.BlobInfo;
 import com.google.cloud.storage.Bucket;
 import com.google.cloud.storage.BucketInfo;
+import com.google.cloud.storage.BucketInfo.LifecycleRule;
+import com.google.cloud.storage.BucketInfo.LifecycleRule.LifecycleAction;
+import com.google.cloud.storage.BucketInfo.LifecycleRule.LifecycleCondition;
 import com.google.cloud.storage.Storage;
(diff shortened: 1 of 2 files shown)


samdgupi:
Please add some details and jira explaining why this change is needed

src/main/java/io/cdap/e2e/utils/StorageClient.java
```diff
@@ -93,4 +98,14 @@ public static Blob uploadObject(String bucketName, String objectName, String fil
       Files.readAllBytes(Paths.get(StorageClient.class.getResource("/" + filePath).toURI())));
   }
 
+  public static Bucket createBucketwithLifeCycle(String bucketName, int age) throws IOException {
```
samdgupi: where is this getting called from ?

samdgupi: please add a java doc. It is unclear what timeunit age is refering to

neerajsinghal05: This is newly created method , which will be called form framework to GCP to enable the lifecycle rule in the bucket with Age.

neerajsinghal05: Java doc has added for the method mentioned above.

samdgupi: Please change the variable name to `ageInDays`

samdgupi (approved):

src/main/java/io/cdap/e2e/utils/StorageClient.java
neerajsinghal05: suggested changes has been done.

PR #233, opened 2023-11-27
add the --test-third-party-cookie-phaseout chrome flag

GnsP:
Launch Chrome using the `--test-third-party-cookie-phaseout` flag for e2e tests.
Ref: [chrome blogpost on third party cookie phaseout](https://developer.chrome.com/en/docs/privacy-sandbox/third-party-cookie-phase-out/#test)

### DIFF
--- a/src/main/java/io/cdap/e2e/utils/SeleniumDriver.java
+++ b/src/main/java/io/cdap/e2e/utils/SeleniumDriver.java
@@ -50,6 +50,7 @@ public class SeleniumDriver {
     chromeOptions.addArguments("--no-sandbox");
     chromeOptions.addArguments("--disable-setuid-sandbox");
     chromeOptions.addArguments("--headless");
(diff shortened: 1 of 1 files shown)


samdgupi:
The blog post mentions running two parallel test runs one with and one without the flag but that would increase the time to test.
LGTM otherwise.

src/main/java/io/cdap/e2e/utils/SeleniumDriver.java
```diff
@@ -50,6 +50,7 @@ public class SeleniumDriver {
     chromeOptions.addArguments("--no-sandbox");
     chromeOptions.addArguments("--disable-setuid-sandbox");
     chromeOptions.addArguments("--headless");
+    chromeOptions.addArguments("--test-third-party-cookie-phaseout");
```
samdgupi, line 53: Please add a jira to remove it once the chrome rollout is done/stabilised.

samdgupi (approved):

PR #234, opened 2023-11-28
Fix BQMT Plugin upload error

Vipinofficial11:
Error - Error in opening the file which uploads the BQMT plugin due to incorrect path.

Fix - Providing exact path to the file to resolve the error.


Vipinofficial11:
After Providing the correct path we are able to upload the plugin, validated if the plugin is present in Artifact list of sandbox or not.

Screenshot 2023-11-29 at 18 18 09

itsankit-google (approved):

Vipinofficial11:
Merging the PR as its approved and validated if plugin is uploading.

PR #237, opened 2023-12-11
Added a script to build CDAP sandbox and reverted the changes from run_e2e_test.py

Vipinofficial11:
1. Added a script to build CDAP sandbox and upload plugins to run e2e tests on latest CDAP code.
2. Reverted the changes made to run_e2e_tests as they are not needed anymore. (They were implemented to run e2e on CDAP only)

Script Testing :  Tested the scripts on a self hosted runner to make sure changes are working.

Screenshot 2023-12-13 at 10 33 31

### DIFF
--- a/src/main/scripts/run_cdap_e2e_test.py
+++ b/src/main/scripts/run_cdap_e2e_test.py
@@ -1,0 +1,140 @@
+# Copyright © 2023 Cask Data, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not
+# use this file except in compliance with the License. You may obtain a copy of
+# the License at
+
+# http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations under
+# the License.
+
+import os
+import subprocess
+import zipfile
+import shutil
+import argparse
+import re
+import configparser
+import sys
+
+
+def run_shell_command(cmd):
+    process = subprocess.run(cmd.split(" "), stderr=subprocess.PIPE)
+    if process.returncode != 0:
+        print("Process completed with error: ", process.stderr)
+    assert process.returncode == 0
+
+
+def get_sandbox_version(directory_path):
+    pattern = re.compile(r"cdap-sandbox-(\d+\.\d+\.\d+)-SNAPSHOT.zip")
+    # Iterate over files and folders in the directory
+    for entry in os.listdir(directory_path):
+        if pattern.match(entry):
+            version_match = pattern.search(entry)
+            if version_match:
+                version = version_match.group(1)
+                return version
+
+
+def get_hydrator_branch(gitmodules_path, submodule_path):
+    config = configparser.ConfigParser()
+    config.read(gitmodules_path)
+
+    # Check if the submodule exists in the .gitmodules file
+    if config.has_section(f"submodule \"{submodule_path}\""):
+        branch = config.get(f"submodule \"{submodule_path}\"", "branch")
+        return branch
+    else:
+        return None
+
+
+# Parse command line optional arguments
+parser = argparse.ArgumentParser()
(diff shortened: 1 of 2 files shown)


src/main/scripts/run_cdap_e2e_test.py
```diff
@@ -0,0 +1,88 @@
...
+subprocess.run(["git", "clone", hydrator_repository_url])
+os.chdir("./hydrator-plugins")
+run_shell_command("git submodule update --init --recursive --remote")
+run_shell_command("mvn clean install -DskipTests")
+os.chdir("../plugin")
+# Building the sandbox
+run_shell_command('mvn clean package -pl cdap-standalone,cdap-app-templates/cdap-etl -am -amd -DskipTests -P '
+                  f'templates,dist,release,unit-tests -Dadditional.artifacts.dir=../../hydrator-plugins/')
+os.chdir("./cdap-standalone/target")
+sandbox_zip = "cdap-sandbox-6.11.0-SNAPSHOT.zip"
```
itsankit-google: Do not hardcode sandbox version, you can simply find it in the target folder.

Vipinofficial11: Added a method to find the version of the sandbox. (ad2e0337b6d5307e65785a24bcc62805b4628d7a)

src/main/scripts/run_cdap_e2e_test.py
```diff
@@ -0,0 +1,101 @@
...
+print("Building CDAP Sandbox")
+os.chdir("./plugin")
+run_shell_command("git submodule update --init --recursive --remote")
+run_shell_command("mvn clean install -DskipTests -P templates,spark-dev")
+my_env = os.environ.copy()
+my_env["MAVEN_OPTS"] = "-Xmx1024m -XX:MaxPermSize=128m"
+my_env["_JAVA_OPTIONS"] = "-Xmx32G"
+# Building the plugins
+os.chdir("..")
+hydrator_repository_url = "https://github.com/cdapio/hydrator-plugins.git"
```
itsankit-google, line 94: If we want to use the same script for release branches, we will need to compute `hydrator-plugins` branch as well from `cdap-build` repo based on cdap branch: https://github.com/cdapio/cdap-build/blob/f785173eb3274b4b9d06e162ce09b9671b2eb091/.gitmodules#L8

Vipinofficial11: Added a logic to get the base branch (ex: develop) from e2e.yaml and then based on it we clone cdap-build repo and the look for hydrator-plugins branch in gitmodules. (0bed67d8331a626ce2a755d88c72c08104428228)

e2e.yaml workflow

```
- name: Run all e2e tests
if: github.event_name == 'workflow_dispatch' || github.event_name == 'push' || steps.filter.outputs.e2e-test == 'true'
 run: python3 e2e/src/main/scripts/run_cdap_e2e_test.py --testRunner TestRunnerRequired.java --cdapBranch "${{ github.event.pull_request.base.ref }}"
```

Testing output :

Screenshot 2023-12-21 at 13 04 54

src/main/scripts/run_cdap_e2e_test.py
```diff
@@ -0,0 +1,134 @@
...
+run_shell_command("mvn clean install -DskipTests -P templates,spark-dev")
+my_env = os.environ.copy()
+my_env["MAVEN_OPTS"] = "-Xmx1024m -XX:MaxPermSize=128m"
+my_env["_JAVA_OPTIONS"] = "-Xmx32G"
+
+# Building the plugins
+os.chdir("..")
+# Extracting hydrator-plugins branch from cdap-build gitmodules.
+cdap_branch = ""
+if args.cdapBranch:
```
itsankit-google, line 73: if `cdapBranch` is not present should we fail the workflow?

itsankit-google (approved):
one comment, otherwise LGTM

src/main/scripts/run_cdap_e2e_test.py
Vipinofficial11: Yes, that makes sense. Added a check to fail the workflow in case the cdapBranch is empty (fdbc4155ceb3c6e2bf31dc78a2352d2c09914397)

Vipinofficial11:
Squashed the commits, merging the PR.

PR #240, opened 2024-01-08
Changed the method in storage client file for multipart feature to be worked with existing 2.3.0 storage client version

rahuldash171:
if we change the code in StorageClient.java from   LifecycleAction.newAbortIncompleteMPUploadAction()
to LifecycleAction.newLifecycleAction("AbortIncompleteMultipartUpload"), to work with existing 2.3.0 storage client version. 
Previously we upgraded the version to 2.8.0 to support gcs Multi part feature but there was a possiblity dependency conflicts . Instead the method was updated in storage client file and it works with the exiting storage client version . Please view the attached screeshot .


rahuldash171:
image

PR #256, opened 2024-12-09
Added step for Run index in pipeline

AnkitCLI:
This step contains Run Index step to wait till visibility of Run 2 index on UI. This PR is related to fix for GcsCreate on cloudbuild.

### DIFF
--- a/src/main/java/io/cdap/e2e/pages/locators/CdfPipelineRunLocators.java
+++ b/src/main/java/io/cdap/e2e/pages/locators/CdfPipelineRunLocators.java
@@ -94,6 +94,9 @@ public class CdfPipelineRunLocators {
   @FindBy(how = How.XPATH, using = "//*[contains(text(),'saved successfully.')]")
   public static WebElement savedSuccessMessage;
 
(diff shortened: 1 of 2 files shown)


src/main/java/io/cdap/e2e/pages/locators/CdfPipelineRunLocators.java
```diff
@@ -94,6 +94,9 @@ public static WebElement locateRecordsInCountOnPluginNodeAsNonZero(String plugin
   @FindBy(how = How.XPATH, using = "//*[contains(text(),'saved successfully.')]")
   public static WebElement savedSuccessMessage;
 
+  @FindBy(how = How.XPATH, using = "//*[@class='run-number' and text()='Run 2 of 2']")
```
itsankit-google, line 97: It is not advisable to use text to find elements. 

Please work with @GnsP to add `data-cy` attributes.

Curious are there 2 runs in one scenario?

AnkitCLI:
Did step changes in the google cloud repo no need to change in e2e-tests so closing this PR

PR #257, opened 2024-12-12
wait step change for selecting Row

AnkitCLI:
This PR contains changes for wait to page load before waiting for text to be present in element value.

### DIFF
--- a/src/main/java/io/cdap/e2e/pages/actions/CdfConnectionActions.java
+++ b/src/main/java/io/cdap/e2e/pages/actions/CdfConnectionActions.java
@@ -129,6 +129,7 @@ public class CdfConnectionActions {
     WaitHelper.waitForElementToBeClickable(CdfConnectionLocators.searchDirectoryInput,
                                            ConstantsUtil.DEFAULT_TIMEOUT_SECONDS);
     ElementHelper.sendKeys(CdfConnectionLocators.searchDirectoryInput, dataRow);
+    WaitHelper.waitForPageToLoad();
     WaitHelper.waitForTextToBePresentInElementValue(CdfConnectionLocators.searchDirectoryInput, dataRow);
     int attempts = 0;
     while (attempts < 5) {


src/main/java/io/cdap/e2e/pages/actions/CdfConnectionActions.java
```diff
@@ -129,7 +129,7 @@ private static void selectConnectionDataRow(String dataRow) {
     WaitHelper.waitForElementToBeClickable(CdfConnectionLocators.searchDirectoryInput,
                                            ConstantsUtil.DEFAULT_TIMEOUT_SECONDS);
     ElementHelper.sendKeys(CdfConnectionLocators.searchDirectoryInput, dataRow);
-    WaitHelper.waitForTextToBePresentInElementValue(CdfConnectionLocators.searchDirectoryInput, dataRow);
+    WaitHelper.waitForPageToLoad();
```
itsankit-google, line 132: should we do both?

```
WaitHelper.waitForPageToLoad();
WaitHelper.waitForTextToBePresentInElementValue(CdfConnectionLocators.searchDirectoryInput, dataRow);
```
@poojantcs wdyt?

poojantcs: Yes, we should.

AnkitCLI: We can use both the steps but right now we are facing error on that particular step in which this step is waiting for tableName to be present in search bar. But it was not showing. So we just remove this step and simply load that page.
embedding_-643155356
As in the above image the tableName was searched but it is not showing in search bar so we did this change.So if the tableName is filtered out then may be we don't need this step to verify table name is present in search bar also.

itsankit-google: so looks like we are not correctly sending the input string in search bar?

Because I can see it in search bar when I do it manually?

image

AnkitCLI: No, we are sending right input in search bar because the table only comes out when we search that table name in search bar. And in the above error image we can see the table name with UUID shown on UI. So till that point it is fine. It is giving error on that wait step.And this step is only failing on github actions neither on CDF nor when we do it locally.

itsankit-google (approved):

PR #285, opened 2025-07-07
Use central-publishing-maven-plugin for deploy.

ritwiksahani:
* Using the central-publishing-maven-plugin to release the changes.
* Distribution management tag is no longer required with this plugin.
* sonatype repository can be removed since we are pulling from central.
* sonatype-snapshots pointed to https://central.sonatype.com/repository/maven-snapshots. This was done as part of the last PR.


itsankit-google (commented):
can you please revert back the changes in github workflow as well?

itsankit-google (approved):

PR #292, opened 2025-08-20
Updating timeouts and data-testid

AnkitCLI:


src/main/java/io/cdap/e2e/utils/ConstantsUtil.java
```diff
@@ -83,15 +83,15 @@ public class ConstantsUtil {
   /**
    * DEFAULT_TIMEOUT_SECONDS: To be used in external wait helpers defined in {@link WaitHelper}
    */
-  public static final int DEFAULT_TIMEOUT_SECONDS = 300;
+  public static final int DEFAULT_TIMEOUT_SECONDS = 420;
```
itsankit-google: I think 5 mins should be enough, where do we see the need to increase it to `420`?

AnkitCLI: changed to 5 mins

itsankit-google (approved):

